Issue #51, opened 2024-09-11
distilbert raises an error


CosimoRulli:
Hi,

thank you for building this amazing repo. My purpose is to train on msmarco a ColBERT model using `distilbert` as backbone.
I took your script [`knowledge_distillation.py`](https://github.com/lightonai/pylate/blob/main/examples/train/knowledge_distillation.py)
and replace `bert-base-uncased` with `distilbert/distilbert-base-uncased`, but I get the following error:
```
  File "/code/examples/train/knowledge_distillation.py", line 40, in <module>
    model = models.ColBERT(model_name_or_path=model_name)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/pylate/lib/python3.12/site-packages/pylate/models/colbert.py", line 232, in __init__
    super(ColBERT, self).__init__(
  File "/home/user/.conda/envs/pylate/lib/python3.12/site-packages/sentence_transformers/SentenceTransformer.py", line 299, in __init__
    modules = self._load_auto_model(
              ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/pylate/lib/python3.12/site-packages/pylate/models/colbert.py", line 1086, in _load_auto_model
    transformer_model = Transformer(
                        ^^^^^^^^^^^^
  File "/home/user/.conda/envs/pylate/lib/python3.12/site-packages/sentence_transformers/models/Transformer.py", line 54, in __init__
    self._load_model(model_name_or_path, config, cache_dir, **model_args)
  File "/home/user/.conda/envs/pylate/lib/python3.12/site-packages/sentence_transformers/models/Transformer.py", line 85, in _load_model
    self.auto_model = AutoModel.from_pretrained(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/pylate/lib/python3.12/site-packages/transformers/models/auto/auto_factory.py", line 564, in from_pretrained
    return model_class.from_pretrained(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.conda/envs/pylate/lib/python3.12/site-packages/transformers/modeling_utils.py", line 3832, in from_pretrained
    model = cls(config, *model_args, **model_kwargs)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #49, opened 2024-09-03
Add dtype flexibility


bclavie:
Hey! Congrats on the release 😄 

My first issue, as promised to @NohTow: scoring is pretty slow, and I think it could be greatly improved by adding extra flexibility, in terms of dtype? Noticeably:
- The model dtype defaults to float32, and even passing in `model_kwargs={"torch_dtype": torch.float16}
` results in the model being fp32, which slows things down a lot for almost no performance improvement, especially on weaker hardware. A separate `.half()` call afterwards is needed.
- Likewise, `retrieve` doesn't have an option to convert the documents fetched from voyager from float32 to float16, and neither does `colbert_scores` or `rerank`. This means we end up needing to do expensive float32 scoring without being able to opt out.
- Due to the above, all scoring, etc, is done with hardcoded float32 that the user cannot control. This is a noticeable slowdown on a few machines that I've tried, and adds virtually no precision (the original colbert library even hardcoded float16 as the default type).

On my machine with very dirty changes, going from the hardcoded float32 to this version took the time to eval on Scifact from ~1.35s/query to 0.85s. I think this is well worth implementing since the complexity isn't gigantic!

**More minor typing flexibility change**
- Voyager is also hardcoded to use fp32. This is less impactful since it doesn't have fp16 support, but users might want to use E4M3 float8, which introduces tolerable degradation in exchange for better storage.

bclavie:
(cc @raphaelsty)

raphaelsty:
Awesome finding @bclavie, if you want to create a branch, or feel free to share some code here and I'll co-author the commit with you :)

bclavie:
Thanks!

My personal branch is a complete mess since it's mostly RAGatouille related, so if you don't mind I can share a few snippets!

In `models/colbert.py`'s `__init__`:
```python
        if self.model_kwargs and self.model_kwargs.get("torch_dtype") == torch.float16 or self.model_kwargs.get("torch_dtype") == 'float16':
            self.half()
```

in `rank.py`:
```python
def rerank(
    documents_ids: list[list[int | str]],
    queries_embeddings: list[list[float | int] | np.ndarray | torch.Tensor],
    documents_embeddings: list[list[float | int] | np.ndarray | torch.Tensor],
    device: str = None,
    fp16: bool = False,
) -> list[list[dict[str, float]]]:
...
if fp16:
    query_embeddings = query_embeddings.astype(np.float16)
    query_documents_embeddings = [
[...]

PR #52, opened 2024-09-11
Loading logic rework

NohTow:
This PR solves some of the issues related to model loading.

First, it removes the model_kwargs `add_pooling_layer` that prevented the instantiation of a pooler within BERT models but is not common across all encoders (see #51).
I considered removing the pooler after initialization, but then the saved PyLate model does not have weights for it and it yields a warning saying that those weights are not properly loaded from checkpoint. Although it does not matter as we are not using it anyways, this message can be misleading. Thus, I choose to let it be, as we are using the sequence_embeddings and not the pooled output, it's a small additional useless computation but I did not find a better solution.

Second, it adds a function to support loading a model created using the stanford-nlp library. This has two benefits:

1. Every ColBERT model (with a base model loadable using ST) is now **natively compatible with PyLate**, without having to convert it manually. This should greatly enhance the number of compatible models (#50).
2. Besides not having to convert it, it also means that we do not have to add the PyLate files to an existing stanford-nlp repository, as we did for Colbert-small. Besides not duplicating the weights, it solves this issue where the Transformer (from ST) folder was not at the root but in a subfolder, which resulted in the model configuration not being properly loaded and thus not properly loading the model to a specified dtype (#49).

Also took the opportunity to add dtype casting to the Dense layer to match the Transformer.

### DIFF
--- a/pylate/models/Dense.py
+++ b/pylate/models/Dense.py
@@ -2,10 +2,12 @@ import json
 import os
 
 import torch
+from safetensors import safe_open
 from safetensors.torch import load_model as load_safetensors_model
 from sentence_transformers.models import Dense as DenseSentenceTransformer
 from sentence_transformers.util import import_from_string
 from torch import nn
+from transformers.utils import cached_file
 
 __all__ = ["Dense"]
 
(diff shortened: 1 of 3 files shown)


raphaelsty (approved):
LGTM

PR #60, opened 2024-10-01
Adding a FastAPI server

NohTow:
This PR adds code to serve the embeddings of a ColBERT model using FastAPI, leveraging [batched](https://github.com/mixedbread-ai/batched) to create dynamic batches.

See the README for details about how it works

### DIFF
--- a/pylate/server/server.py
+++ b/pylate/server/server.py
@@ -12,12 +12,36 @@ app = FastAPI()
 
 
 class EmbeddingRequest(BaseModel):
+    """PyDantic model for the requests sent to the server.
+
+    Parameters
+    ----------
+    input
+        The input(s) to encode.
+    is_query
+        A boolean indicating if the input is a query or a document.
+    model
+        The name model to use for encoding.
+    """
+
     input: List[str] | str
     is_query: bool = True
     model: str = "lightonai/colbertv2.0"
 
 
 class EmbeddingResponse(BaseModel):
+    """PyDantic model for the server answer to a call.
+
+    Parameters
+    ----------
+    data
+        A list of dictionaries containing the embeddings ("embedding" key) and the type of the object ("object" key, is always embedding).
+    model
+        The name of the model used for encoding.
+    usage
+        An approximation of the number of tokens used to generate the embeddings (computed by splitting the input sequences on spaces).
+    """
+
     data: List[dict]
     model: str
     usage: dict
--- a/pylate/server/server.py
+++ b/pylate/server/server.py
@@ -86,6 +86,13 @@ model.encode_document = batched.aio.dynamically(
 
 @app.post("/v1/embeddings", response_model=EmbeddingResponse)
 async def create_embedding(request: EmbeddingRequest):
+    """API endpoint that encode the elements of an EmbeddingRequest and returns an EmbeddingResponse.
+
+    Parameters
+    ----------
+    request
+        The EmbeddingRequest containing the elements to encode, the model to use, and whether the input is a query or a document.
+    """
     if request.model != args.model:
         raise HTTPException(
             status_code=400,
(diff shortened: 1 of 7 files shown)


pylate/server/server.py
```diff
@@ -0,0 +1,99 @@
...
+import uvicorn
+from fastapi import FastAPI, HTTPException
+from pydantic import BaseModel
+
+from pylate import models
+
+app = FastAPI()
+
+
+class EmbeddingRequest(BaseModel):
```
raphaelsty, line 14: Would be great to add some docstring here

pylate/server/server.py
```diff
@@ -0,0 +1,99 @@
...
+    input: List[str] | str
+    is_query: bool = True
+    model: str = "lightonai/colbertv2.0"
+
+
+class EmbeddingResponse(BaseModel):
+    data: List[dict]
```
raphaelsty, line 21: Here as well

pylate/server/server.py
```diff
@@ -0,0 +1,99 @@
...
+        default="lightonai/colbertv2.0",
+        help="Model to serve, can be an HF model or a path to a model",
+    )
+    return parser.parse_args()
+
+
+args = parse_args()
+
+# We need to load the model here so it is shared for every request
+model = models.ColBERT(args.model)
```
raphaelsty, line 55: We should be loading the model at startup only ?

pylate/server/server.py
```diff
@@ -0,0 +1,99 @@
...
+# We cannot create the function on the fly as the batching require to use the same function (memory address)
+model.encode_query = batched.aio.dynamically(wrap_encode_function(model, is_query=True))
+model.encode_document = batched.aio.dynamically(
+    wrap_encode_function(model, is_query=False)
+)
+
+
+@app.post("/v1/embeddings", response_model=EmbeddingResponse)
+async def create_embedding(request: EmbeddingRequest):
```
raphaelsty, line 64: You can add a docstring

pylate/server/server.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def wrap_encode_function(model, **kwargs):
+    def wrapped_encode(sentences):
+        return model.encode(sentences, **kwargs)
+
+    return wrapped_encode
+
+
+def parse_args():
```
raphaelsty, line 33: We could add a cli, but parse_args look good to me

pylate/server/server.py
raphaelsty: On this specific point, I think it would be better to create a class 

```python
class Model:
    """""""
    def __init__(self) ->:
        """""
        self.model = None

    def encode(self, text) ->:
         """""
         return self.model.encode(text)

    def load(self, model) ->:
        self.model = models.ColBERT(args.model)
```

Initialize the model just like you initialize the ColBERT model and in the dedicated endpoint which is launched at startup by fastapi execute `model.load(model)` otherwise you will load the model at each call no ?

raphaelsty: useful resource: https://fastapi.tiangolo.com/advanced/events/#lifespan

raphaelsty:
LGTM some details related to formatting, the loading is more serious as we should find a way to load the model only once 😀

pylate/server/server.py
raphaelsty: Closing this as the model is only loaded once

pylate/server/server.py
NohTow: Done

pylate/server/server.py
NohTow: Done

NohTow:
Merging as the only remaining factor was the model being loaded multiple time while it is not the case.

PR #62, opened 2024-10-09
Check if architectures is in the config before reading it

NohTow:
When experimenting around, I realized that some models do not have "architectures" in their config (e.g, [debertav3](https://huggingface.co/microsoft/deberta-v3-base).

This PR just check if there is one before checking if it is a HF_ColBERT (and thus that the model is a stanford-nlp model).


raphaelsty:
LGTM

PR #63, opened 2024-10-11
Make knowledge distillation processing compatible with DictDataset and Dataset.

raphaelsty:
This MR update the knowledge distillation processing to make it compatible with DictDataset and Dataset. While the previous typing showed that this module take as input Dataset, it does not. It takes only as input DictDataset. Make this module compatible with both DictDataset and Dataset. Enable also the user to define the split which is defined to "train" by default just like the previous version.

### DIFF
--- a/pylate/utils/processing.py
+++ b/pylate/utils/processing.py
@@ -15,7 +15,9 @@ class KDProcessing:
         Queries dataset.
     documents
         Documents dataset.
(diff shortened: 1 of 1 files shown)


NohTow:
Good catch, did not encounter this as my datasets did not contain the word "train" lmao

Issue #64, opened 2024-10-14
Loading SentenceTransformer Model and `_load_sbert_model` in ColBERT class


BenoitHardier:
Hey, thanks for your work. 
I am trying to finetuned a model based on `OrdalieTech/Solon-embeddings-base-0.1`. 
like that:
```
 model = models.ColBERT(model_name_or_path="OrdalieTech/Solon-embeddings-base-0.1")
```

Unfortunatly, it seems that the `_load_sbert_model` should output both `modules` and `module_kwargs` (The SentenceTranformer doc seems wrong) .
I am not sure if it's related to the SentenceTransformer version but I am using the last one 3.2.0

I finally succeed to load solon changing the `_load_sbert_model` in `colbert.py`
It should be modified:
https://github.com/lightonai/pylate/blob/ddaf8f82dfd8725275c2362c2560c78565b5a874/pylate/models/colbert.py#L1153 and here https://github.com/lightonai/pylate/blob/ddaf8f82dfd8725275c2362c2560c78565b5a874/pylate/models/colbert.py#L1194-L1199

Let me know, if I am doing something wrong or if my issue is useful.

Benoît HARDIER

NohTow:
Hello,

Indeed, breaking changes were introduced in [this commit of ST](https://github.com/UKPLab/sentence-transformers/commit/6257cb00ef87ddc8f445f0e436377ea4c48879b2), hence why we fixed the version of sentence-transformers to the last version before 3.1 (sentence-transformers == 3.0.1).

The fix is indeed what you suggested and during my preliminary tests, it seemed to work fine, but I wanted to make sure there was no regression before doing a PR for the fix and bump the version of ST. I might do the PR in the next days if it becomes needed.
In the mean time, you can either use your own hacked version or use the 3.0.1 version if you do not need the latest change (e.g, ONNX backend if you are training on CPU)!

NohTow:
Hello again,

I fixed the issue in #65 and now the training with the model from Ordalie works fine with ST 3.2.0!
I'll merge the PR during the day alongside with other improvements and bump the version of PyLate.
![image](https://github.com/user-attachments/assets/90c3ac5b-0644-4c7f-ab21-e01895b77a4e)

BenoitHardier:
Hey

Great, It's also working on my setup.

I will try to continue digging into the repo. (I am specially interested in training/fine-tuning french retrieval models)
I let you close the issue when you have merged the fix. 

Benoît HARDIER

[...]

PR #65, opened 2024-10-15
Fixing issue with ST 3.1 and bumping version

NohTow:
This PR fixes a compatibility issue introduced in ST 3.1 in [this commit](https://github.com/UKPLab/sentence-transformers/commit/6257cb00ef87ddc8f445f0e436377ea4c48879b2).

Basically, this added the possibility add module_kwargs in the config so that the model can leverage them for a specific processing, such as the dataset_embeddings of [cde](https://huggingface.co/jxm/cde-small-v1).

This fix allows to bump the version of ST to newer ones and also to handle such custom processing aswell (as we are extending the ST model).

Edit for more information: the [load_sbert_model function set the self.modules_kwargs](https://github.com/UKPLab/sentence-transformers/blob/a4be00f3fcb635f536566044d40c41513a495818/sentence_transformers/SentenceTransformer.py#L306) that are then used in the [forward pass of the Transformer model](https://github.com/UKPLab/sentence-transformers/blob/1802076d4eae42ff0a5629e1b04e75785d4e193b/sentence_transformers/SentenceTransformer.py#L686) if some are [set in the config.json](https://huggingface.co/jxm/cde-small-v1/blob/main/modules.json#L7)

Related issue: https://github.com/lightonai/pylate/issues/64

### DIFF
--- a/pylate/models/colbert.py
+++ b/pylate/models/colbert.py
@@ -301,7 +301,7 @@ class ColBERT(SentenceTransformer):
             )
             self[1] = Dense.from_sentence_transformers(dense=self[1])
         else:
-            logger.warning("Pylate model loaded successfully.")
+            logger.warning("PyLate model loaded successfully.")
 
         # Ensure all tensors in the model are of the same dtype as the first tensor
         try:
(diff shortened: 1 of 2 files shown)


raphaelsty:
LGTM

Issue #59, opened 2024-09-30
Set default to non normalization of the scores


NohTow:
Following JaColBERTv2.5, I added the [normalization of the scores for distillation.](https://github.com/lightonai/pylate/blob/d526d98b1b44899c2ba7e2c20addb22be2faecbd/pylate/losses/distillation.py#L112)

Yet, this setup is not [really standard/well-know](https://github.com/lightonai/pylate/discussions/58) and so the default should be no normalization, with a proper documentation on using normalization and its potential benefits. This would avoid confusion and possible sub-optimal results if the user does not normalize the teacher scores on its end aswell.

PR #66, opened 2024-10-15
[Draft] Setting normalize_scores default to False

NohTow:
This PR change the default of normalize_scores to False and add documentation about the parameters.
The rationale behind this change is two-fold:
1. The setup is not really common for the large audience and [the user can be surprised by the loss magnitude](https://github.com/lightonai/pylate/discussions/58). Also, if the user use his own dataset without normalizing the teacher scores, it could be hurting the performance.
2. [JinaColBERT-v2 had mitigated results using normalization](https://arxiv.org/pdf/2408.16672), so I am not sure it is worth it to make it the default.

Before merging, there is two things that needs to be discussed:
1. The datasets we have released have normalized scores, which could make the boilerplate a bit off.
2. When building the functionality, I decoupled the normalization of the output scores and the teachers' to offload part of the computation pre-training as it would be re-usable. But now that I think about it again, it might be more painful for the user and we could add the normalization of the teacher scores during the training aswell.

Related issue: https://github.com/lightonai/pylate/issues/59

### DIFF
--- a/docs/documentation/training.md
+++ b/docs/documentation/training.md
@@ -176,6 +176,12 @@ trainer.train()
 
     Refer to this [documentation](https://sbert.net/docs/sentence_transformer/training/distributed.html) for more information.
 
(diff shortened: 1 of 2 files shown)


raphaelsty:
@NohTow Merging this MR

PR #67, opened 2024-10-15
Creation of the PylateModelCard

NohTow:
This PR fixes the issue with the model_card creation mentioned in #58 by creating a dedicated PylateModelCard.
Besides fixing the issue, it also allows to create a card adapted to PyLate models while also having the original useful information for ST cards.

Let's first assert that everything is correct in the card before merging.

Edit: the failing test is due to the version of ST being too old. I'll rebase on main once we merge the PR for bumping the version.


pylate/hf_hub/model_card.py
```diff
@@ -0,0 +1,259 @@
...
+
+    versions["tokenizers"] = tokenizers_version
+    return versions
+
+
+@dataclass
+class PylateModelCardData(SentenceTransformerModelCardData):
+    """A dataclass storing data used in the model card.
+
+    Args:
```
raphaelsty, line 64: Could you update the docstring to match our docstring template (it will be used in the doc)

pylate/models/colbert.py
```diff
@@ -217,14 +214,15 @@ def __init__(
         model_kwargs: dict | None = None,
         tokenizer_kwargs: dict | None = None,
         config_kwargs: dict | None = None,
-        model_card_data: Optional[SentenceTransformerModelCardData] = None,
+        model_card_data: Optional[PylateModelCardData] = None,
```
raphaelsty, line 217: you can use  `PylateModelCardData | None = None` here for consistency

NohTow: Done

raphaelsty:
LGTBM

tomaarsen:
Very exciting!

PR #69, opened 2024-10-28
Stanford loading using bin file

NohTow:
#68 fixed the local loading of Stanford models, but the issue is that the safe tensor is only created when uploaded to HF through the dedicated bot.
This PR changes the loading logic to rely on the `pytorch_model.bin` file that is always there both locally and remotely.
cc @jessiejuachon

### DIFF
--- a/pylate/models/Dense.py
+++ b/pylate/models/Dense.py
@@ -2,7 +2,6 @@ import json
 import os
 
 import torch
-from safetensors import safe_open
 from safetensors.torch import load_model as load_safetensors_model
 from sentence_transformers.models import Dense as DenseSentenceTransformer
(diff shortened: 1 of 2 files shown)


NohTow:
Tests failing are not related to the PR but to [this](https://github.com/UKPLab/sentence-transformers/pull/3026).
We should either wait for ST to create the version with the merged fixes or set the version of transformers to pre-4.46.

NohTow:
I updated the version of ST with the fixes and pinned the version of transformers to avoid future breaking changes.
We can now merge @raphaelsty.

raphaelsty:
Amazing ! Go merge :D

PR #73, opened 2024-11-28
Add safetensor OR bin loading logic + add loading tests

NohTow:
#69 actually broke the loading of recent stanford-nlp models (small/jina), as they do not have bin.
That is on me, I checked the colbertv2 repository that had both safe tensor AND bin and rushed to the conclusion that all of them had both.

This PR merge old safe tensor loading and the new bin loading to be able to load both type of repository.
We first try to load safe tensors and fall back to bin if not existing.

Also added some tests for loading different types of models (base, ST base, stanford-nlp bin/safetensor and PyLate), as well as a small check that should make sure that the model is not only loaded but correctly loaded.

### DIFF
--- a/pylate/models/Dense.py
+++ b/pylate/models/Dense.py
@@ -123,6 +123,7 @@ class Dense(DenseSentenceTransformer):
                     use_auth_token=use_auth_token,
                 )
             except EnvironmentError:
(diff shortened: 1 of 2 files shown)


raphaelsty:
LGTM, in the next versions we might want to expand those tests

PR #85, opened 2025-01-30
ref: ruff python 3.8 

sam-hey:
Align Ruff with setup.py: Set Python version to **3.8** for Ruff to ensure consistency with `setup.py` 

https://github.com/lightonai/pylate/blob/c6a22fcdc313ee1c2ced454f77bd4c199cb964b4/setup.py#L59

### DIFF
--- a/docs/documentation/datasets.md
+++ b/docs/documentation/datasets.md
@@ -119,7 +119,7 @@ Example entry:
 }
 ```
 
-> [!WARNING]  
+???+ warning
     Ensure that the length of `document_ids` matches the length of `scores`.
 
 ### Queries
--- a/docs/documentation/datasets.md
+++ b/docs/documentation/datasets.md
@@ -119,7 +119,7 @@ Example entry:
 }
 ```
 
-???+ warning
+> [!WARNING]  
     Ensure that the length of `document_ids` matches the length of `scores`.
 
 ### Queries
--- a/docs/documentation/evaluation.md
+++ b/docs/documentation/evaluation.md
@@ -95,10 +95,12 @@ The output is a dictionary containing various evaluation metrics. Here’s a sam
 ```
 
 
-> [!NOTE]
-> 1. is_query flag: Always set is_query=True when encoding queries and is_query=False when encoding documents. This ensures the model applies the correct prefixes for queries and documents.
-> 2. Evaluation metrics: The pipeline supports a wide range of evaluation metrics, including NDCG, hits, MAP, recall, and precision, with different cutoff points.
-> 3. Relevance judgments (qrels): The qrels are used to calculate how well the retrieved documents match the ground truth.
+???+ info
+    1. is_query flag: Always set is_query=True when encoding queries and is_query=False when encoding documents. This ensures the model applies the correct prefixes for queries and documents.
+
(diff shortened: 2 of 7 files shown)


NohTow:
About the warning message: we not only load the configuration of the stanford-nlp model, but also the weights.
Actually, we more load the weights than the config, as I wanted to load more information from the model config by reading [artificat.metada](https://huggingface.co/colbert-ir/colbertv2.0/blob/main/artifact.metadata), such as the max doc/query length.

sam-hey:
I applied your feedback to make the warning message more precise and also fixed some documentation.

NohTow:
I am nitpicking but the new message does not highlight the fact that the _model_ has been loaded. Why do you want to emphasize configuration (especially as right now we do not really load anything from the config, just use the default markers)? 
I will make the PR to actually load information from stanford-nlp configurations so it will be more accurate, but to me this is secondary: we load the model, which include the config.
Maybe something along: 
```
"Loaded the ColBERT model from Stanford NLP. Please ensure that the configuration aligns with your expected setup."
```

sam-hey:
I want to highlight this for two main reasons: First, I encountered an issue with the configuration not being loaded properly, and second, I wasn’t sure what exactly was being loaded that could be affiliated with `Stanford NLP`. Specifically, I wasn’t certain whether the local checkpoint I specified had been properly loaded, or if there was an error when passing the local path to the library, causing it to load **something** from the Stanford NLP Project to fill in the gap instead.

Since the event loaded isn’t a problem, I was quite irritated by the warning. It might be better to split this up for better clarity.

```python
logger.info("Loaded the ColBERT model.")
logger.warning("If no custom prefixes are set, the default Stanford NLP prefixes will be used. Please ensure that the configuration aligns with your expected setup.")

```

What do you think about this change?

NohTow:
Hello,
To give a better fix to the issue you encountered when using stanford models, I opened #88, that actually reads the values from stanford metadata to set the parameters. As for every model, if the user feed those parameters, it overrides the config that is read, but loading Stanford NLP model without any additional args will now load markers, lengths and attend to expansion token correctly. I also split the warning into the loading of the weights and the loading of the config, with a different warning if the config is not read properly.

Could you have a look and tell me if this solves your issue or if you want other parameters to be read?
Thanks!

sam-hey:
A huge thank you from my side! I did add two comments, but they're just small nitpicks :) Otherwise, I tested it locally, and it looks good to me. Thanks a lot for the effort!

So just the Ruff version in this PR :)

NohTow:
You're very welcome, it was a long time overdue that I procrastinated for too long, should have been done when Jina-ColBERT came out with different parameters than the defaults!
> I did add two comments

Where did you add those? I am sorry I think I missed them

> So just the Ruff version in this PR :)

You also added some changes to the documentation at some point, but I am not sure I understood those because the documentation layout looks fine to me?

sam-hey:
I added them as a review, thinking you would be able to see them, but it seems they are in a pending state, so that’s not the case. I also added a comment, which should be visible to you.
> You also added some changes to the documentation at some point, but I am not sure I understood those because the documentation layout looks fine to me?

I removed them because I was reviewing the files on GitHub, where warnings and other markers were broken, as the documentation on the website uses different markers. I didn’t realize that at the time.

PR #87, opened 2025-02-04
Bumping transformers and st versions

NohTow:
This PR bumps transformers and sentence-transformers versions.
Transformers bump allows the use of ModernBERT.
Sentence-transformers allows to fix [an issue](https://github.com/UKPLab/sentence-transformers/pull/3110) to load PyLate model trained with `trust_remote_code=True` that was introduced in a modification of the code of ST back then.

This is a draft because I identified some breaking changes that I'll fix.


NohTow:
Well it seems that the issue I had identified was the only one (minus the less verbose logging of NanoBEIR now).
The only change I had to make was to make [set_widget_examples a no-op](https://github.com/lightonai/pylate/pull/87/commits/c0a15bc39dab211b78ba98dbe482102ee42e240e) because there is issues with it and our data mapping/transform. I already did a modification to skip its call back then, but Tom reintroduced a call to it, so the best thing to do is just make it a no-op for now, except if you want to debug.

Tests are passing and I have been training with transformers 4.48.2 for some time, as well as a new version of ST (not the final one though) and had no issue. As tests are passing, I guess it is safe to merge @raphaelsty

NohTow:
I fixed a small issue where the model cards created when training on top of a ST models were just copy of the existing model card instead of the new one, due to a change in ST.
I actually removed the overriding of the `create_model_card function` because it was useless and was not updated due to this.

pylate/hf_hub/model_card.py
```diff
@@ -231,25 +228,21 @@ def to_dict(self) -> dict[str, Any]:
...
-            super_dict["similarity_fn_name"] = "Cosine Similarity"
-
+            super_dict["similarity_fn_name"] = "MaxSim"
         self.first_save = False
 
         for key in IGNORED_FIELDS:
             super_dict.pop(key, None)
         return super_dict
 
+    # For now, set_widget_examples is not compatible with our transform/map operations, so we make it a no-op until it is fixed
```
raphaelsty, line 242: You can add this comment to the docstring of the function instead of putting it before

PR #88, opened 2025-02-04
Read stanford configs

NohTow:
Right now, we are using default values for doc/query length, markers and attending to expansion tokens when reading stanford models.
This causes some issues as highlighted in #85 and also requires the user to specify a lot of information when using a model that is not using default values, as can be seen with the [loading of Jina-ColBERT.](https://lightonai.github.io/pylate/models/models/#available-models)

This PR simply add the reading process of the `artifact.metadata` file of Stanford NLP models and read markers, lengths and attend to expansion tokens values.

As usual, we override those if the user feed values to the init of the model. Also changed the `attend_to_expansion_tokens` parameter to match the other (None by default and override at the end).

### DIFF
--- a/pylate/models/colbert.py
+++ b/pylate/models/colbert.py
@@ -19,6 +19,7 @@ from sentence_transformers.quantization import quantize_embeddings
 from sentence_transformers.util import batch_to_device, load_file_path
 from torch import nn
 from tqdm.autonotebook import trange
+from transformers.utils import cached_file
 
 from ..hf_hub.model_card import PylateModelCardData
 from ..scores import SimilarityFunction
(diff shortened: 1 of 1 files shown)


NohTow:
I think it should be ok @raphaelsty 
Could you please review to make sure the behavior is in line with what's expected? I did some tests, but I might have missed something

sam-hey:
It seems that my comments are in Pending state, so they can't be seen at the moment. I have just one small regrade:

I wouldn’t categorize this as a warning, as it’s expected behavior. It’s merely informing you that the StanfordNLP model has successfully loaded the weights.
https://github.com/lightonai/pylate/blob/f0899ab5468f451ee46fe5559514fd791eaab97e/pylate/models/colbert.py#L268
Similarly, I believe there should be a clear distinction between informational events and actual warnings. For instance, if there were an issue loading the file and it fell back to a default, that would merit a warning.


https://github.com/lightonai/pylate/blob/f0899ab5468f451ee46fe5559514fd791eaab97e/pylate/models/colbert.py#L294-L303

NohTow:
That is very fair, I don't know why I started using `logger.warning`, I think I used it for a proper warning once and then copied it over the whole loading logic, my bad.
I'll clean that up before merging, thanks for the remark!

NohTow:
@sam-hey I changed some warnings to info
@raphaelsty you can merge after you have double checked the loading/overriding logic

raphaelsty:
LGTM

PR #92, opened 2025-02-23
fix: use model card from pylate for  upload to hub

sam-hey:
This PR allows the use of push_to_hub with the PyLate template. 

ST lib is importing its own create ModelCard function that is not getting overwritten by PyLate. 

```

from pylate import models
from pylate.models.colbert import PylateModelCardData

model = models.ColBERT(
    model_name_or_path="colbert-ir/colbertv2.0",
    model_card_data=PylateModelCardData(model_name="GerColBERT2"),
)

model.push_to_hub(".....", private=True,  exist_ok=True)



```

### DIFF
--- a/pylate/models/colbert.py
+++ b/pylate/models/colbert.py
@@ -1213,12 +1213,6 @@ class ColBERT(SentenceTransformer):
         # we don't generate a new model card, but reuse the old one instead.
         if self._model_card_text and "generated_from_trainer" not in self.model_card_data.tags:
             model_card = self._model_card_text
-            if self.model_card_data.model_id:
-                # If the original model card was saved without a model_id, we replace the model_id with the new model_id
-                model_card = model_card.replace(
-                    'model = SentenceTransformer("sentence_transformers_model_id"',
-                    f'model = SentenceTransformer("{self.model_card_data.model_id}"',
-                )
         else:
             try:
                 model_card = generate_model_card(self)
--- a/pylate/models/colbert.py
+++ b/pylate/models/colbert.py
@@ -6,6 +6,8 @@ import logging
 import math
 import os
 import string
+import traceback
+from pathlib import Path
 from typing import Iterable, Literal, Optional
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


sam-hey:
If you're generally okay with this PR, @NohTow, what do you think about testing this? We could address it by pushing to a private LightOnAI hf repo and then loading from it as a test case. Does that sound like a good idea to you?

NohTow:
Hello,
First, sorry if we are delaying you a bit lately, lot of things on my plate!


I actually did not test pushing to hub the models trained, but the model card created locally during the saving seemed to be overwritten by our template, so the save function looks alright?
If I check the [`push_to_hub` function from ST](https://github.com/UKPLab/sentence-transformers/blob/309a9accb784e0931636ad97be7806e37c57dfb4/sentence_transformers/SentenceTransformer.py#L1432), it just pushes the folder to the hub, so it should be fine.
Do the files saved locally during your training have the correct model card or not?

I am trying to figure out what's wrong, let me know if I am missing something

NohTow:
Actually I think I understand, the card was working before #87, where I removed the function to not override a function that has no change w.r.t ST and benefits from Tom's changes without doing anything, but there is a regression.
Could you do the `push_to_hub` of a folder created after training (possibly with NanoBEIR evaluation) and give me a link just so I double check? Otherwise it seems fine to me

sam-hey:
No worries about the delay!

I've submitted a PR for ST to improve this, as I share the same sentiment. ^^

I'll keep this PR open as an alternative in case it's rejected. If it's accepted, I'll repurpose this PR to integrate PyLate with my changes.

https://github.com/UKPLab/sentence-transformers/pull/3253

pylate/models/colbert.py
```diff
@@ -1186,3 +1188,47 @@ def _load_sbert_model(
...
+            if not model_path.exists() and not self.model_card_data.model_id:
+                self.model_card_data.model_id = model_name
+
+        # If we loaded a Sentence Transformer model from the Hub, and no training was done, then
+        # we don't generate a new model card, but reuse the old one instead.
+        if self._model_card_text and "generated_from_trainer" not in self.model_card_data.tags:
+            model_card = self._model_card_text
+            if self.model_card_data.model_id:
+                # If the original model card was saved without a model_id, we replace the model_id with the new model_id
+                model_card = model_card.replace(
```
NohTow, line 1218: If we keep this function, we need to change this no?

sam-hey: Yes, definitely 🫣

sam-hey:
The changes have been merged into ST! Thanks a lot for your support, @NohTow! Looking forward to ST v4+ so we can bring this update to PyLate.

Have a great weekend!

PR #94, opened 2025-02-28
fix:  Loading Model - Stanford Models

sam-hey:
@raphaelsty @NohTow 

I believe this is an urgent fix that also requires a new release.  
```python
from pylate import models


model = models.ColBERT(
    model_name_or_path="colbert-ir/colbertv2.0",
)
```

```
No sentence-transformers model found with name colbert-ir/colbertv2.0.
Traceback (most recent call last):
  File "/Users/A200009373/Documents/Coding/pylate/test.py", line 7, in <module>
    model = models.ColBERT(
            ^^^^^^^^^^^^^^^
  File "/Users/A200009373/Documents/Coding/pylate/pylate/models/colbert.py", line 283, in __init__
    self.query_prefix = metadata["query_token_id"]
                        ~~~~~~~~^^^^^^^^^^^^^^^^^^
KeyError: 'query_token_id'
```

### DIFF
--- a/pylate/models/colbert.py
+++ b/pylate/models/colbert.py
@@ -279,18 +279,27 @@ class ColBERT(SentenceTransformer):
                     with open(metadata, "r") as f:
                         metadata = json.load(f)
                         # If the user do not override the values, read from config file
-                        if self.query_prefix is None:
-                            self.query_prefix = metadata["query_token_id"]
-                        if self.document_prefix is None:
-                            self.document_prefix = metadata["doc_token_id"]
-                        if self.query_length is None:
-                            self.query_length = metadata["query_maxlen"]
-                        if self.document_length is None:
-                            self.document_length = metadata["doc_maxlen"]
-                        if self.attend_to_expansion_tokens is None:
-                            self.attend_to_expansion_tokens = metadata[
-                                "attend_to_mask_tokens"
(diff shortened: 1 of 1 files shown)


raphaelsty:
We need to increase the coverage @NohTow, releasing a new version right now.

We need dedicated test with distinct kind of models

NohTow:
Deeply sorry about that, I once again over-generalized the structure of Stanford models, for my defense this time I checked multiples models, but I guess this was only added after a while...

In any case, sorry again I should always add some checks even if it seems to be coherent and also add some tests when I add features, my bad!
Thanks a lot for spotting this @sam-hey!

sam-hey:
Thanks for the quick response and fix!

raphaelsty:
@sam-hey The version 1.1.6 is available thank you to have spotted the issue. We will add a set of unit test to cover those issue. Amazing that it take you 10 min after the release to spot the issue 😅🙏

sam-hey:
I've been eagerly awaiting this release! : ) I wanted to uncomment some tests I wrote for MTEB, tested with PyLate installed from source, and quickly ran into this problem - all the previously passed tests suddenly started failing locally.

So thanks again for your effort and quick response! Have a nice weekend!

PR #98, opened 2025-03-06
Adding GradCache (cached contrastive loss)

NohTow:
This PR adds CachedContrastive loss to the available loss.
As described in the [paper](https://arxiv.org/pdf/2101.06983), this effectively allows to scale the batch size of contrastive loss without requiring more memory (at the cost of a slightly slower training) by decomposing the loss computation into representations computation without the gradient to compute similarities over the whole batch and then computing the gradients on sub batches.

The code is vastly copied from sentence-transformers implementation (thanks @tomaarsen), one of the main difference being the chunking of the score computation as well (as MaxSim is pretty memory intensive for very large batches).

### DIFF
--- a/pylate/losses/cached_contrastive.py
+++ b/pylate/losses/cached_contrastive.py
@@ -90,6 +90,32 @@ class CachedContrastive(nn.Module):
         Whether to average or sum the cross-entropy loss across the mini-batch.
     show_progress_bar : bool
         Whether to show a TQDM progress bar for the embedding steps.
+    Examples
+    --------
+    >>> from pylate import models, losses
+
+    >>> model = models.ColBERT(
+    ...     model_name_or_path="sentence-transformers/all-MiniLM-L6-v2", device="cpu"
+    ... )
+
+    >>> loss = losses.CachedContrastive(model=model, mini_batch_size=1)
+
+    >>> anchors = model.tokenize([
+    ...     "fruits are healthy.", "chips are not healthy."
+    ... ], is_query=True)
+
+    >>> positives = model.tokenize([
+    ...     "fruits are good for health.", "chips are not good for health."
+    ... ], is_query=False)
+
+    >>> negatives = model.tokenize([
+    ...     "fruits are bad for health.", "chips are good for health."
+    ... ], is_query=False)
+
+    >>> sentence_features = [anchors, positives, negatives]
+
+    >>> loss = loss(sentence_features=sentence_features)
+    >>> assert isinstance(loss.item(), float)
     """
 
     def __init__(
--- a/pylate/losses/cached_contrastive.py
+++ b/pylate/losses/cached_contrastive.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 from collections.abc import Iterator
 from contextlib import nullcontext
 from functools import partial
-from typing import Iterable, Optional
+from typing import Callable, Iterable, Optional
 
 import torch
 import torch.nn.functional as F
@@ -79,13 +79,13 @@ class CachedContrastive(nn.Module):
     ----------
     model :
         A PyLate ColBERT model
-    score_metric : Callable
+    score_metric
         ColBERT scoring function. Defaults to colbert_scores.
-    mini_batch_size : int
+    mini_batch_size
         Chunk size for the forward pass. You can keep this small to avoid OOM on large batch sizes.
-    size_average : bool
+    size_average
         Whether to average or sum the cross-entropy loss across the mini-batch.
-    show_progress_bar : bool
+    show_progress_bar
         Whether to show a TQDM progress bar for the embedding steps.
 
     Examples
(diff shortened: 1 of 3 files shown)


pylate/losses/cached_contrastive.py
```diff
@@ -0,0 +1,348 @@
...
+        A PyLate ColBERT model
+    score_metric : Callable
+        ColBERT scoring function. Defaults to colbert_scores.
+    mini_batch_size : int
+        Chunk size for the forward pass. You can keep this small to avoid OOM on large batch sizes.
+    size_average : bool
+        Whether to average or sum the cross-entropy loss across the mini-batch.
+    show_progress_bar : bool
+        Whether to show a TQDM progress bar for the embedding steps.
+    Examples
```
raphaelsty, line 93: Add space between arguments definitions and unit tests examples

pylate/losses/cached_contrastive.py
```diff
@@ -0,0 +1,348 @@
...
+        loss = self.calculate_loss(reps, masks, with_backward=True)
+        loss = loss.detach().requires_grad_()
+
+        self.cache = [
+            [r.grad for r in rs] for rs in reps
+        ]  # e.g. 3 * bsz/mbsz * (mbsz, hdim)
+
+        return loss
+
+    def calculate_loss(self, reps, masks, with_backward: bool = False) -> Tensor:
```
raphaelsty, line 220: Add typing and docstring with arguments definition so it's easier to follow

raphaelsty:
LGTM, this feature is amazing and was not straightforward to adapt

Note, we must be careful with the docstrings otherwise they won't be properly parsed by the system. Here is the standard we defined where the function / class method definition starts on the same line as the quotes. There is a space between the set of parameters and examples

```python
"""Computes the ColBERT scores between queries and documents embeddings. The score is computed as the sum of maximum similarities
    between the query and the document.

    Parameters
    ----------
    queries_embeddings
        The first tensor. The queries embeddings. Shape: (batch_size, num tokens queries, embedding_size)
    documents_embeddings
        The second tensor. The documents embeddings. Shape: (batch_size, num tokens documents, embedding_size)

    Examples
    --------
    >>> import torch

    >>> queries_embeddings = torch.tensor([
    ...     [[1.], [0.], [0.], [0.]],
    ...     [[0.], [2.], [0.], [0.]],
    ...     [[0.], [0.], [3.], [0.]],
    ... ])

    >>> documents_embeddings = torch.tensor([
    ...     [[10.], [0.], [1.]],
    ...     [[0.], [100.], [1.]],
    ...     [[1.], [0.], [1000.]],
    ... ])

    >>> scores = colbert_scores(
    ...     queries_embeddings=queries_embeddings,
    ...     documents_embeddings=documents_embeddings
    ... )

    >>> scores
    tensor([[  10.,  100., 1000.],
            [  20.,  200., 2000.],
            [  30.,  300., 3000.]])

    """
```

pylate/losses/cached_contrastive.py
```diff
@@ -0,0 +1,348 @@
...
+        self._fork.__exit__(exc_type, exc_val, exc_tb)
+        self._fork = None
+
+
+def _backward_hook(
+    grad_output: Tensor,
+    sentence_features: Iterable[dict[str, Tensor]],
+    loss_obj,
+) -> None:
+    """
```
raphaelsty, line 47: Description should start on the same line as the quote """ for documentation parsing

pylate/losses/cached_contrastive.py
```diff
@@ -0,0 +1,348 @@
...
+                # Dot the embedding chunk with the cached gradient chunk,
+                # scaled by grad_output from the top-level backward pass
+                surrogate = (
+                    torch.dot(reps_mb.flatten(), grad_mb.flatten()) * grad_output
+                )
+                surrogate.backward()
+
+
+class CachedContrastive(nn.Module):
+    """
```
raphaelsty, line 75: Description should start on the same line as the quote """ for documentation parsing

pylate/losses/cached_contrastive.py
```diff
@@ -0,0 +1,348 @@
...
+
+        self.cache = [
+            [r.grad for r in rs] for rs in reps
+        ]  # e.g. 3 * bsz/mbsz * (mbsz, hdim)
+
+        return loss
+
+    def calculate_loss(self, reps, masks, with_backward: bool = False) -> Tensor:
+        """Calculate the cross-entropy loss. No need to cache the gradients."""
+        # Each sub-list in reps is a list of mini-batch chunk embeddings
```
raphaelsty, line 222: You could add those comments to the docstring or to the parameters definitions so it's clear from our documentation, the user would not have to dig the code to find those informations

pylate/losses/cached_contrastive.py
```diff
@@ -0,0 +1,348 @@
...
+from torch import Tensor, nn
+from torch.utils.checkpoint import get_device_states, set_device_states
+
+from ..models import ColBERT
+from ..scores import colbert_scores
+from .contrastive import extract_skiplist_mask
+
+
+class RandContext:
+    """
```
raphaelsty, line 20: Description should start on the same line as the quote """ for documentation parsing

raphaelsty:
LGTM merging 🥳🐐😎

PR #100, opened 2025-03-19
Add gathering to (cached) contrastive loss

NohTow:
This PR adds the option to gather documents from the other GPUs in both contrastive and cached contrastive loss.
The default is still not to gather even though in my opinion it should be the default for back-compatibility purpose.
Also, this is most likely bringing the most improvement when all the GPUs have batches from the same dataset in the multi dataset setting. 
I am having discussion with @tomaarsen to define what is the best way to include this option into ST, but for now I have found a workaround and will add a boilerplate later.

The sanity check I ran on a ms marco dataset:
- GradCache is almost equivalent to standard contrastive for the same bs (Not == because I cannot seed everything into ST)
- Increasing the global batch size with grad cache while using the old batch size as mini batch size gets better perf without using more VRAM
- Gathering brings a lot of improvements and is compatible works with both grad cache and contrastive loss
- Increasing the global batch size with grad cache and gathering yield the best results without increasing VRAM usage

Also enhanced a bit the documentation.

### DIFF
--- a/pylate/utils/distributed.py
+++ b/pylate/utils/distributed.py
@@ -1,0 +1,14 @@
+import torch.distributed as dist
+from typing import Sequence
+import torch
+def all_gather(tensor: torch.Tensor) -> Sequence[torch.Tensor]:
+    """Collects a Tensor from each rank, and returns a list of gathered Tensors indexed by rank. The tensor for the local rank is the original one, with the gradients while the others have no gradients.
+    """
+    if dist.is_available() and dist.is_initialized():
+        obj_gather_list = [torch.zeros_like(tensor) for _ in range(dist.get_world_size())]
+        dist.all_gather(obj_gather_list, tensor)
+        obj_gather_list[dist.get_rank()] = tensor
+        return obj_gather_list
+    world_size = dist.get_world_size()
+    if world_size == 1:
+        return [tensor]
--- a/pylate/utils/distributed.py
+++ b/pylate/utils/distributed.py
@@ -1,14 +1,25 @@
 import torch.distributed as dist
 from typing import Sequence
 import torch
+import logging
+
+not_init_warning = True
+logger = logging.getLogger(__name__)
 def all_gather(tensor: torch.Tensor) -> Sequence[torch.Tensor]:
     """Collects a Tensor from each rank, and returns a list of gathered Tensors indexed by rank. The tensor for the local rank is the original one, with the gradients while the others have no gradients.
     """
+    global not_init_warning
     if dist.is_available() and dist.is_initialized():
         obj_gather_list = [torch.zeros_like(tensor) for _ in range(dist.get_world_size())]
         dist.all_gather(obj_gather_list, tensor)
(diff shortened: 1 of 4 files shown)


pylate/utils/distributed.py
```diff
@@ -0,0 +1,33 @@
...
+
+import logging
+from typing import Sequence
+
+import torch
+import torch.distributed as dist
+
+not_init_warning = True
+logger = logging.getLogger(__name__)
+
```
raphaelsty, line 11: The following version is cleaner

```python
from __future__ import annotations

import logging
from typing import Sequence

import torch
import torch.distributed as dist

logger = logging.getLogger(__name__)
_has_warned_dist_not_initialized = False


def all_gather(tensor: torch.Tensor) -> Sequence[torch.Tensor]:
    """Gathers a tensor from each distributed rank into a list.
    
    - If torch.distributed is available and initialized:
      1. Creates a list of tensors (each sized like the input `tensor`).
      2. Gathers tensors from each rank into that list.
      3. Replaces the local tensor in the list with the original tensor that
         retains gradients.

    - If torch.distributed is either unavailable, uninitialized, or
      `world_size == 1`, it returns a singleton list containing only the
      original tensor. In that case, it emits a warning message the first
      time this code path is reached (helpful when using a single GPU setup).

    Parameters
    -----------
        tensor (torch.Tensor): The input tensor to be gathered from each rank.

    Returns
    --------
        Sequence[torch.Tensor]: A list of tensors collected from each rank. On
        a single GPU or when distributed is uninitialized, the list will contain
        only the original tensor.
    """
    global _has_warned_dist_not_initialized

    # Check if torch.distributed is properly available and initialized.
    if dist.is_available() and dist.is_initialized():
        world_size = dist.get_world_size()
        gathered_tensors = [torch.zeros_like(tensor) for _ in range(world_size)]
        
        # Perform all_gather.
        dist.all_gather(gathered_tensors, tensor)
        
        # Replace local rank's tensor with the original (retaining gradients).
        local_rank = dist.get_rank()
        gathered_tensors[local_rank] = tensor
        return gathered_tensors

    # If distributed is not available or initialized, try retrieving world_size.
    try:
        world_size = dist.get_world_size()
        # If world_size == 1, just return the original tensor in a list.
        if world_size == 1:
            return [tensor]
    except ValueError:
        # We get here if dist.get_world_size() is called while not initialized.
        pass

    # Warn once about uninitialized or single-GPU usage.
    if not _has_warned_dist_not_initialized:
        logger.warning(
            "Attempting to gather a tensor while torch.distributed is not "
            "available or not initialized. Returning the local tensor only. "
            "This warning will appear only once."
        )
        _has_warned_dist_not_initialized = True

    return [tensor]
```

raphaelsty (commented):
LGMT, you only need to update `pylate/utils/distributed.py` which is not clear, otherwise everything is great. Good idea to create a dedicated file to gather tensors.

raphaelsty (approved):
LGTM

PR #103, opened 2025-03-24
Update setup.cfg

Samoed:
Pylate can't be install with setuptools >78. Ref https://github.com/pypa/setuptools/issues/4910

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,2 @@
 [metadata]
-description-file = README.md
+description_file = README.md


NohTow:
Not very familiar with the setuptools issue, but if it is back-compatible I don't see why we should not merge this.
That being said, I assignated @raphaelsty because we he is better suited to have the pyproject/setup discussion, but maybe we want to have this discussion in the dedicated issue and not this PR.

Samoed:
The issue with `setup.cfg` is that it shouldn't contain keys with dashes, as they have been deprecated since 2021 (https://github.com/pypa/setuptools/issues/4910#issuecomment-2748489811) and won't be supported after 2026 (https://github.com/pypa/setuptools/pull/4911/files#r2010652906)

sam-hey:
I can confirm that this change is necessary with the latest version of pip (25.0.1). I had to apply this change to successfully install pylate.

NohTow:
cc @raphaelsty

PR #111, opened 2025-04-11
Scale contrastive loss by world size

NohTow:
With the recent addition of gathering from other gpus in a multi-gpu setting, there's a scaling by the world size missing to the loss (because PyTorch does an average over the number of GPUs by default, so we need to scale by the same number to have a similar loss no matter the number of GPUs).
Also took the opportunity to create a dedicated function for getting the rank instead of duplicating code.

I wonder if we should set the world size/rank as a parameter to only read it once instead of at each step... Maybe it's too much optimization idk

### DIFF
--- a/pylate/utils/distributed.py
+++ b/pylate/utils/distributed.py
@@ -111,3 +111,12 @@ def get_rank() -> int:
         return dist.get_rank()
     else:
         return 0
+
+
+def get_world_size() -> int:
+    """Returns the world size in a distributed training."""
+    # Check if torch.distributed is properly available and initialized.
+    if dist.is_available() and dist.is_initialized():
+        return dist.get_world_size()
+    else:
+        return 1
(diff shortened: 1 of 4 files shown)


pylate/utils/distributed.py
```diff
@@ -102,3 +102,21 @@ def all_gather_with_gradients(tensor: torch.Tensor) -> Sequence[torch.Tensor]:
...
 
     return [tensor]
+
+
+def get_rank() -> int:
+    """Returns the current rank in a distributed training."""
+    # Check if torch.distributed is properly available and initialized.
+    if dist.is_available() and dist.is_initialized():
+        return dist.get_rank()
+    else:
```
raphaelsty, line 112: remove the else and simply return 0

NohTow:
@raphaelsty I removed the else and directly return even though I find the explicit way more readable, but it's a matter of test IG
I suppose we can merge after the tests run

raphaelsty:
@NohTow LGTM, once test pass we can merge 🚀

PR #115, opened 2025-05-05
Add temperature parameter to contrastive losses

NohTow:
Hey,
This PR adds a temperature parameter to the contrastive losses.
I initially removed it when copying the losses from ST because it is named "scale" and was actually a multiplicative scalar instead of being called temperature and dividing the similarities.
It turns out, for contrastive training, we require a very low temperature if we want to get the best performance possible.
I set the default to 1.0 to be equivalent to previous trainings, but maybe using 0.02 is better as it is the most commonly used value.
WDYT @raphaelsty?

### DIFF
--- a/pylate/losses/cached_contrastive.py
+++ b/pylate/losses/cached_contrastive.py
@@ -132,6 +132,7 @@ class CachedContrastive(nn.Module):
         size_average: bool = True,
         gather_across_devices: bool = False,
         show_progress_bar: bool = False,
+        temperature: float = 1.0,
     ) -> None:
         super(CachedContrastive, self).__init__()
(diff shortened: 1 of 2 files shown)


raphaelsty:
Amazing MR as it brings a lot of value, I would put the value which allow to produce the best-in-class model possible, so 0.02 🙌

NohTow:
I am not sure, this has been shown very effective, but I think forcing it as a default might be a bit confusing for the user.
Maybe we should let it default to 1.0 and add more information about it in the documentation.

NohTow:
Merging like that, we'll add information about the parameter and its optimal value in documentation.

PR #121, opened 2025-05-13
Fix: Handle index loading correctly to prevent stream read error

rajj0208:
This fixes a bug where loading an existing index file caused a RuntimeError due to incorrect use of Index.load. The code was passing a file path instead of a file object. The fix opens the file in binary mode and passes the file handle to Index.load, resolving the issue.

### DIFF
--- a/pylate/indexes/voyager.py
+++ b/pylate/indexes/voyager.py
@@ -162,7 +162,8 @@ class Voyager(Base):
 
         """
         if os.path.exists(path=index_path) and not override:
-            return Index.load(index_path)
+            with open(index_path, "rb") as f:
+                return Index.load(f)
 
         if os.path.exists(path=index_path):
             os.remove(index_path)


NohTow:
Hey!
Thanks for contributing to PyLate.
It was surprising me because I did not change much to the loading of the index lately and we had no issue, so I tried a small test where I index nfcorpus, then load the index and do the evaluation.
I do not have any error when loading the index, could you please provide information about your tests/a reproducible example so I can see when this happens?
Also, could you provide information about your environment? Maybe it has to do with Python version or something.

[The Voyager documentation seems to say that path is supported (although so is file_like)](https://spotify.github.io/voyager/python/reference.html#voyager.Index.load)

rajj0208:
Hi,
Thanks for the quick response!

You're right, the Voyager documentation does state that both file paths and file-like objects are supported when loading the index. After some investigation, I found that the issue only occurs in my local environment (running via VS Code). When I run the same code in a cloud environment like Google Colab, it works without any issues using Index.load(file_path).

To be thorough, I tested this across multiple Python environments and versions locally (including Python 3.9, 3.10 and 3.12), and the issue persists. However, when I use with open(file_path, "rb") and pass the file-like object instead, the index loads fine locally.

It seems like the difference might stem from how the local environment resolves or interacts with file paths, or perhaps how the Voyager bindings behave across environments. 

Try to run the code after storing the index locally.

NohTow:
I've been running those on my mac as well as on different remote nodes without issue.
I am not against doing a modification to the loading, but I would love to understand what causes the issue so we can better fix it (because it might be that file-like might fail on other setup then)
What is your local setup?

rajj0208:
Here's my local setup where the issue occurs:

OS: Windows 11
Python Versions Tested: 3.9.18, 3.10.13, 3.12.9 (issue occurs across versions)
Environment: Tested in conda environments
IDE: VS Code

[https://lightonai.github.io/pylate/documentation/retrieval/](Running this simple code to load index)

NohTow:
> OS: Windows 11

I think it might be the cause of the issue, Windows probably do not handle filepath as Linux does there. 
You are not using WSL right?
Could you print `index_path` type in both Windows and Linux to check if there is a difference?

I'll be making some tests to see if file-like works well on Linux, if it does, I don't see any reason to use something that works across more plateforms! I'll also be widening the test of the indexes to include saving/loading and be able to catch such issues.

rajj0208:
No, I wasn't using WSL, I was using the regular Command Prompt on Windows. I’ve checked the type of `index_path` on both Windows and Linux, and it's `<class 'str'>` in both cases. However, the format of the path is different, with Windows using backslashes (\) and Linux using forward slashes (/).

I agree that it would be helpful to test with the file-like approach on Linux to see if it resolves the issue. Yes, you can perform with your tests, and if this works well, I think switching to a more platform-agnostic solution would be beneficial.

NohTow:
Yes, definitely!
Do you mind if I do the modification/tests a bit later? It is not blocking you right now right?

We have the PLAID MR coming up and I have other stuff to do first, if that is ok.

NohTow:
Well the PLAID branch could have been merged today (almost) but we are having issue with CI idling somehow.
If the fix you suggest works out of the box for all setup, I can add the tests quickly and merge after rebasing, so it should not take long if we fix CI quickly
If somehow we struggle too long I'll do this PR first, but I would prefer not if possible

rajj0208:
Got it! Just to be clear, I don’t think I have the merge permissions, so I’ll leave that to you.

NohTow:
I fixed CI, we should merge PLAID today so I can have a look at this!
Is this ok with you if I create a new branch and add you as a co-author of the PR?
Would make my life easier rather than pushing on the fork (not sure I am allowed to)

> Just to be clear, I don’t think I have the merge permissions, so I’ll leave that to you.

Yes I know, was more oriented towards Raphaël if he saw the PR in the mean time :)

rajj0208:
Sounds great! I’m perfectly fine with you creating a new branch and adding me as a **co-author** of the PR. That approach seems like a good way to streamline the process, especially with the CI and merge flow. Looking forward to seeing the PR merge soon.

rajj0208:
Hi, have you tested whether it's working across all platforms? Also, when do you plan to merge #123?

NohTow:
We just merged PLAID and extended its support to Windows.
I am currently doing the tests for Voyager and extending our coverage in #123 as you can see.

NohTow:
Closing as solved in #123.

PR #123, opened 2025-05-15
Use file-like loading for the Voyager index to fix Windows loading (and add loading tests for indexes)

NohTow:
This is #121 locally to extend the branch and add some tests.
Essentially, Voyager do not handle index loading from path on Windows (I suppose they do not added the case), but file-like loading is working and it seems to also work on Linux.


NohTow:
@raphaelsty I added some loading tests/addition/deletion to Voyager/PLAID and it seems to run fine on Mac/Linux/Windows
Also took the opportunity to move all files from indexes to a common subdirectory as discussed. It breaks existing indexes but it's for the best in the future.
Also removed an useless mention to checkpoint in PLAID.

PR #127, opened 2025-05-20
fix: update pylate img path for pypi

sam-hey:
https://pypi.org/project/pylate/

PR #128, opened 2025-05-20
ci: run tests on pull request & push

sam-hey:
And right away, another one! Thanks a lot!

- Run tests on pull request and on push

### DIFF
--- a/.github/workflows/plaid-tests.yml
+++ b/.github/workflows/plaid-tests.yml
@@ -1,5 +1,11 @@
 name: PLAID Tests
-on: [push]
+on:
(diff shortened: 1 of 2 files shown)


sam-hey:
Hi @raphaelsty, I don't think so—have a look at the earlier PR:
image
https://github.com/lightonai/pylate/pull/127
 
 Only the linting was triggered because I just opened the PR. In this case, the push rule doesn’t apply. Since you're working on a branch within the repository, it counts as both opening a PR and pushing to a branch inside the project, which leads to different behavior.

PR #134, opened 2025-06-24
Allow higher transformers version

carlesonielfa:
Newer models can't be used because of the strict transformers version.

Related to #133.

- Matched the transformers version specifier to that of sentence transformers.
- Ran tests locally and all are passing without issue

Issue #133, opened 2025-06-18
Update transformers dependency to >=4.51.0


marcofiocco:
...otherwise I cannot use Qwen3 as dense embeddings. 
I'm using Qdrant vector db with Qwen3:8b dense embeddings, so I can use Reason-ModernColBERT for reranking.
My assumption is that Qwen3 is state of the art and faster (is it?) than Reason-ModernColBERT  for vector search. 
Please advice.

NohTow:
Hey,

Sorry for the delay, was on vacation!
I did a branch to allow for broader version range of transformers on this [branch](https://github.com/lightonai/pylate/compare/main...range_transformers_version).
As ST now supports 5.0, maybe we can also increase the upper bound to 5.0 but I need to check that everything works fine.

Qwen3 has better bench results than most existing models, that being said, me and others have had mixed results with it, it should still be a strong model though. About speed, it is a single vector dense model, so it might different than multi-vector in terms of speed, but Qwen3 is a 8B model while ModernColBERT is 150M, so I am not sure at all it will be faster.
Finally, please note that Reason-ModernColBERT is meant for specific data (reasoning-intensive data). If you want to use a multi-vector model for "generic" data (what is found in MTEB, and what Qwen3 is trained for), you should use [GTE-ModernColBERT](https://huggingface.co/lightonai/GTE-ModernColBERT-v1).

marcofiocco:
Thanks.
interesting. Do you think it makes sense to use Reason-ModernColBERT for vector search AND also for reranking in the same pipline or is it redundant?

NohTow:
Should be fixed in #137!

PR #137, opened 2025-07-01
transformers version range

NohTow:
Hey,
This PR change the transformers requirement from a pinned version to a range, to allow use of PyLate with different setups (notably for RAGatouille). 
Also update the max version of transformers to allow the use of Qwen 3.
Related to #133 and #134


NohTow:
ping @raphaelsty

ssubu (approved):

Issue #139, opened 2025-07-25
Error using Readme example with eval steps


FelixLabelle:
### **Bug**

When using the example from the readme and setting eval_steps, the trainer fails to evaluate over the dev set.

**Env Info**
Windows 11
Python  3.12
Using pylate installed from git
Needed to tweak some packages to get this running
- Triton Windows port v 3.4
- Pytorch 2.8.1 + Cuda 12.8
- Transformers 4.54.0.dev (needed the latest code for models I'm interested in running)

### **Code:**
```
import torch
from datasets import load_dataset
from sentence_transformers import (
    SentenceTransformerTrainer,
    SentenceTransformerTrainingArguments,
)

from pylate import evaluation, losses, models, utils

# Define model parameters for contrastive training
model_name = "bert-base-uncased"  # Choose the pre-trained model you want to use as base
batch_size = 32  # Larger batch size often improves results, but requires more memory

num_train_epochs = 1  # Adjust based on your requirements
# Set the run name for logging and output directory
run_name = "contrastive-bert-base-uncased"
output_dir = f"output/{run_name}"

# 1. Here we define our ColBERT model. If not a ColBERT model, will add a linear layer to the base encoder.
model = models.ColBERT(model_name_or_path=model_name)

# Compiling the model makes the training faster
model = torch.compile(model)

# Load dataset
dataset = load_dataset("sentence-transformers/msmarco-bm25", "triplet", split="train")
# Split the dataset (this dataset does not have a validation set, so we split the training set)
splits = dataset.train_test_split(test_size=0.01)
train_dataset = splits["train"]
eval_dataset = splits["test"]

# Define the loss function
train_loss = losses.Contrastive(model=model)

# Initialize the evaluator
dev_evaluator = evaluation.ColBERTTripletEvaluator(
    anchors=eval_dataset["query"],
    positives=eval_dataset["positive"],
    negatives=eval_dataset["negative"],
)

# Configure the training arguments (e.g., batch size, evaluation strategy, logging steps)
args = SentenceTransformerTrainingArguments(
    output_dir=output_dir,
    num_train_epochs=num_train_epochs,
    per_device_train_batch_size=batch_size,
    per_device_eval_batch_size=batch_size,
    fp16=True,  # Set to False if you get an error that your GPU can't run on FP16
    bf16=False,  # Set to True if you have a GPU that supports BF16
    run_name=run_name,  # Will be used in W&B if `wandb` is installed
    learning_rate=3e-6,
[...]

PR #143, opened 2025-08-29
Cast indexing numpy int to Python int

NohTow:
Hello,

As mentioned in #139, an update of `datasets` broke the encoding function (which is the reason why tests are failing although the same code was passing before).
As in [ST](https://github.com/UKPLab/sentence-transformers/pull/3455), the simple fix is just to cast the numpy int to a python integer.


raphaelsty:
LGTM

PR #146, opened 2025-09-06
feat: optimize skiplist_mask with O(1) lookup table approach

jacobloveless:
--- TL; DR ---
- Replace O(n×m) algorithm with O(1) LUT-based implementation
- Add device-aware caching for skiplist lookup tables
- Achieve 77.7x speedup on GPU, 33.2x on CPU
- Include comprehensive tests and benchmarks
- Maintain full backward compatibility

--- Full Detail ---
# Optimize skiplist_mask with O(1) lookup table approach - 77.7x speedup on GPU

## Summary

This PR optimizes the `skiplist_mask` method in ColBERT by replacing the O(n×m) algorithm with an O(1) lookup table approach, resulting in **77.7x speedup on GPU** and **33.2x speedup on CPU**.

## Problem

The current `skiplist_mask` implementation has O(n×m) time complexity where:
- n = number of input tokens
- m = size of skiplist (punctuation/special tokens to mask)

For each token in the skiplist, it performs a `torch.where` operation across all input tokens, creating m intermediate tensors and executing m comparison operations.

## Solution

This PR introduces a lookup table (LUT) based approach with O(1) complexity per token:

1. **Build once**: Create a boolean tensor of size `vocab_size` where skiplist tokens are marked as False
2. **Cache**: Store the LUT per (device, skiplist) combination to avoid rebuilding
3. **Apply**: Use simple tensor indexing `lut[input_ids]` for masking

### Key Implementation Details

- **Device-aware caching**: Separate cache entries for CPU and each GPU device
- **Immutable cache keys**: Skiplist converted to sorted tuple for consistent caching
- **Dynamic vocab sizing**: Automatically determines vocabulary size with 50k minimum
- **Memory efficient**: Typical BERT vocab (30k) uses only ~30KB per cache entry

## Performance Results

Benchmarked on NVIDIA L4 GPU with typical ColBERT workloads:

### Speed Improvements

| Configuration | Original (ms) | Optimized (ms) | Speedup |
|--------------|---------------|----------------|---------|
| Batch=8, Seq=128, Skip=50 | 2.506 | 0.033 | **75.9x** |
| Batch=32, Seq=512, Skip=100 | 4.900 | 0.040 | **122.5x** |
| **Average across all tests** | - | - | **77.7x** |

### Cache Effectiveness

- First call (cache miss): 0.165ms
- Subsequent calls (cache hit): 0.040ms
- Cache speedup: **4.1x**

### Visualization

The PR includes comprehensive benchmarks showing:
- Consistent speedup across different batch sizes
- Linear scaling with sequence length
- Minimal performance degradation with larger skiplists

## Correctness

All outputs are identical to the original implementation, verified through:
- Unit tests with edge cases (empty skiplist, single token, all tokens)
- Integration testing with actual ColBERT models
- Bit-exact comparison of mask outputs

## Code Quality

- **Extensive documentation**: 40+ lines of comments explaining the algorithm, performance characteristics, and implementation details
- **Type hints**: Full type annotations for clarity
- **Error handling**: Assertions for input validation
- **Clean code**: Following PyLate coding standards

## Testing

A comprehensive benchmark script is included (`tests/test_skiplist_optimization.py`) that:
- Compares original vs optimized performance
- Tests various batch sizes, sequence lengths, and skiplist sizes
- Verifies correctness across edge cases
- Measures cache effectiveness
- Analyzes memory usage

To run the benchmark:
```bash
python tests/test_skiplist_optimization.py
```

## Impact

This optimization significantly improves ColBERT training and inference performance:
- **Training**: Faster forward passes reduce epoch time
- **Inference**: Lower latency for real-time applications
- **Scaling**: Better GPU utilization for production deployments

## Backward Compatibility

The optimization is fully backward compatible:
- Same input/output interface
- Same behavior and results
- Cache is transparent to users

## Memory Considerations

The LUT cache uses minimal memory:
- ~30KB per unique skiplist (for BERT-sized vocab)
- Automatic cleanup when ColBERT instance is deleted
- Bounded by number of unique skiplist combinations (typically <10)

## References

- Original issue: Performance bottleneck identified in production ColBERT deployments
- Related work: Similar optimizations in other transformer implementations

---

**Note to reviewers**: The dramatic speedup comes from eliminating the nested loop structure. Instead of m passes over n tokens (O(n×m)), we now do a single indexing operation (O(n)) with a pre-computed lookup table.

### DIFF
--- a/PR_DESCRIPTION.md
+++ b/PR_DESCRIPTION.md
@@ -1,0 +1,111 @@
+# Optimize skiplist_mask with O(1) lookup table approach - 77.7x speedup on GPU
+
+## Summary
+
+This PR optimizes the `skiplist_mask` method in ColBERT by replacing the O(n×m) algorithm with an O(1) lookup table approach, resulting in **77.7x speedup on GPU** and **33.2x speedup on CPU**.
+
+## Problem
+
+The current `skiplist_mask` implementation has O(n×m) time complexity where:
+- n = number of input tokens
+- m = size of skiplist (punctuation/special tokens to mask)
+
+For each token in the skiplist, it performs a `torch.where` operation across all input tokens, creating m intermediate tensors and executing m comparison operations.
+
+## Solution
+
+This PR introduces a lookup table (LUT) based approach with O(1) complexity per token:
+
+1. **Build once**: Create a boolean tensor of size `vocab_size` where skiplist tokens are marked as False
+2. **Cache**: Store the LUT per (device, skiplist) combination to avoid rebuilding
+3. **Apply**: Use simple tensor indexing `lut[input_ids]` for masking
+
+### Key Implementation Details
+
+- **Device-aware caching**: Separate cache entries for CPU and each GPU device
+- **Immutable cache keys**: Skiplist converted to sorted tuple for consistent caching
+- **Dynamic vocab sizing**: Automatically determines vocabulary size with 50k minimum
+- **Memory efficient**: Typical BERT vocab (30k) uses only ~30KB per cache entry
+
+## Performance Results
+
+Benchmarked on NVIDIA L4 GPU with typical ColBERT workloads:
+
+### Speed Improvements
+
+| Configuration | Original (ms) | Optimized (ms) | Speedup |
+|--------------|---------------|----------------|---------|
+| Batch=8, Seq=128, Skip=50 | 2.506 | 0.033 | **75.9x** |
+| Batch=32, Seq=512, Skip=100 | 4.900 | 0.040 | **122.5x** |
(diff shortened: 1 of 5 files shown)


NohTow:
Hey,
Thanks for the PR!

From my understand, the main speed-up should be thanks to apply masking through the indexing rather than the loop (sorry about the loop btw, should have coded it more cleanly in the first place, was in a rush for the first release back then).
Is the LUT really beneficial? I feel like the skiplist is fixed, at least for a given instantiated model so we could store only a single list (as we are already doing with words/ids right now) and use it for indexing

Am I missing something?

PR #151, opened 2025-09-17
[DRAFT] ST 5.0

NohTow:
This PR bumps the ST dependency to the new major version of ST.
This was mostly a version oriented towards sparse models, so it does not seems like there was a lot of breaking changes besides these small ones.
I trained some models and did not encounter any issue for now.


raphaelsty:
Amazing, LGTM

Issue #119, opened 2025-05-12
Cannot `pip install pylate` due to dependency on voyager


grossjohann:
`pylate` seems to say it depends on `voyager>=2.0.9` but voyager is already included in `pylate`.  `pip install pylate` fails.

I am on an M2 Mac running macOS Sequoia 15.4.1 (24E263).  I used Homebrew to install pyenv.  Then I used pyenv to install Python 3.13.3.  Then I ran `pip install pylate` which failed.  To get cleaner output, I repeated `nohup pip install pylate > pip-install.log` and I'm attaching the file `pip-install.log` here.

[pip-install.log](https://github.com/user-attachments/files/20157522/pip-install.log)

NohTow:
Yeah as @raphaelsty the logs seem a bit odd, because it just says that you require `Voyager>=2.0.9`, which is true but should not be an issue. A "conflict" is mentioned, but from the log we cannot really tell why there is any conflict.

grossjohann:
> Yeah as [@raphaelsty](https://github.com/raphaelsty) the logs seem a bit odd, because it just says that you require `Voyager>=2.0.9`, which is true but should not be an issue.

On my system, it doesn't find `Voyager`.  I tried `pip install Voyager` and `pip install voyager`, and both responded: ERROR: Could not find a version that satisfies the requirement Voyager (from versions: none)

> A "conflict" is mentioned, but from the log we cannot really tell why there is any conflict.

Oh yes, I forgot.  I had already tested this but forgot to show the results of that.  See above.

grossjohann:
> Could you try first to uninstall voyager

It wasn't installed...

> And then could you try to install PyLate from source maybe ?

I was able to make it work by installing PyLate ignoring dependencies, what was the command line option... _searches help output_  Ah, `--no-deps`.  Of course, my little test script crapped out at every step but it was pretty obvious what to install.

It is possible that this worked only because I didn't need voyager, I only wanted to compute embeddings from sentences.

But I do have a file `site-packages/pylate/indexes/voyager.py`, it's only 350 lines or so.  I can't tell whether that means there is Python magic at work “do a lot with little code” or whether the file is just an interface to the real functionality.

NohTow:
> On my system, it doesn't find Voyager. I tried pip install Voyager and pip install voyager, and both responded: ERROR: Could not find a version that satisfies the requirement Voyager (from versions: none)

[...]

PR #172, opened 2025-10-28
Move voyager to an optional dependency

NohTow:
As Voyager is not supported for Python 3.12 and the default index is now PLAID/Fast-Plaid, we are moving the Voyager dependency to an optional one to allow PyLate to run on newer version of Python, see #119

### DIFF
--- a/pylate/indexes/voyager.py
+++ b/pylate/indexes/voyager.py
@@ -6,7 +6,6 @@ import os
 import numpy as np
 import torch
 from sqlitedict import SqliteDict
-from voyager import Index, Space
 
 from ..utils import iter_batch
 from .base import Base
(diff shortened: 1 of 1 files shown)


pylate/indexes/voyager.py
```diff
@@ -144,6 +143,9 @@ def _create_collection(
             Whether to override the collection if it already exists.
 
         """
+        # We import voyager here because it is now an optional dependency
+        from voyager import Index, Space
```
Samoed: I think you can add like
```suggestion
        try:
            from voyager import Index, Space
        except ImportError:
            raise ImportError(
                "To use voyager index, please install it with `pip install pylate[voyager]`"
            )
```

PR #176, opened 2025-10-28
Add docs ci

Samoed:
Build docs automatically on pushes to main. Put `docs/api` under ignore, because it will be generated automatically


raphaelsty:
Thank you @Samoed, LGTM

Issue #180, opened 2025-11-26
Error while indexing large datasets like webis-touche using Fast Plaid


meetdoshi90:
Error message on torch quantile with FastPLAID
I was trying to evaluate web-touche using the evaluation code provided by pylate (1.3.4). I am using FastPLAID index with torch 2.8.0. Any wordarounds on this would be appreciated. Works fine for other smaller datasets.
Related issues: [torch-64947](https://github.com/pytorch/pytorch/issues/64947)

> thread '<unnamed>' panicked at /root/.cache/puccinialin/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tch-0.20.0/src/wrappers/tensor_generated.rs:14691:57:
[...]

PR #181, opened 2025-11-28
Fix for #180

meetdoshi90:
This PR proposes a fix for #180 by replacing the existing torch.quantile implementation with one based on torch.kthvalue. Because torch.quantile limits tensors to 16 million elements, it causes problems when computing codebooks for more than 50,000 vectors of dimension greater than 335. 

The new implementation does not have that limitation. For linear interpolation, it is 2x slower on GPU and 10x slower on CPU, but it has been verified to produce correct results. Any suggestions to speed up the CPU implementation are welcome.

A corresponding implementation in Rust would require more effort since it would mean replacing the underlying ATen kernel for quantile.

### Runtime comparison

```
Params
Vector dims: (50000, 768) and (50000, 384)
Quantiles: [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]
Reduction dim: 0
```

Tensor dim = (50000, 768)

| Interpolation | Torch (CPU) | This implementation (CPU) | Torch (GPU) | This implementation (GPU) |
| ------------: | ----------: | ------------------------: | ----------: | ------------------------: |
|        linear |    0.4742 s |                  4.9824 s |    0.0908 s |                  0.2105 s |
|         lower |    0.4488 s |                  2.4526 s |    0.0090 s |                  0.0829 s |
|        higher |    0.4555 s |                  2.4452 s |    0.0090 s |                  0.0835 s |
|      midpoint |    0.4431 s |                  4.8587 s |    0.0092 s |                  0.1697 s |
|       nearest |    0.4648 s |                  2.4603 s |    0.0090 s |                  0.0831 s |

Tensor dim = (50000, 384)

| Interpolation | Torch (CPU) | This implementation (CPU) | Torch (GPU) | This implementation (GPU) |
| ------------: | ----------: | ------------------------: | ----------: | ------------------------: |
|        linear |    0.2071 s |                  3.0143 s |    0.0887 s |                  0.1025 s |
|         lower |    0.2115 s |                  1.5261 s |    0.0052 s |                  0.0273 s |
|        higher |    0.2098 s |                  1.5005 s |    0.0051 s |                  0.0271 s |
|      midpoint |    0.2095 s |                  3.0220 s |    0.0054 s |                  0.0600 s |
|       nearest |    0.2073 s |                  1.4914 s |    0.0051 s |                  0.0272 s |

### DIFF
--- a/pylate/indexes/stanford_nlp/utils/torch_utils.py
+++ b/pylate/indexes/stanford_nlp/utils/torch_utils.py
@@ -2,7 +2,7 @@
 Torch limits quantile operation to 2^24 elements, this can be problematic for colbert indexing with larger embedding dims while computing the codebooks.
 """
 
-from typing import Optional
+from typing import Optional, Union
 
 import torch
 from torch import Tensor
@@ -10,7 +10,7 @@ from torch import Tensor
 
 def torch_quantile(
     input: Tensor,
-    q: float | Tensor,
+    q: Union[float, Tensor],
     dim: Optional[int] = None,
     keepdim: bool = False,
     *,


NohTow:
@meetdoshi90 I'm sorry we're taking time to merge this.
How painful is it for you to add a check about the size (dim * sampled) and goes one route or another?
If this is ok with you, I would prefer this because the og PLAID index has the purpose of being the legacy one and we should keep a way to match old setups

PR #185, opened 2025-12-22
Add tests for 3.13 and 3.14 python

Samoed:
Now this can't be run, because voyager in `dev`. Maybe use it only `voyager` group or test it seperatly?

### DIFF
--- a/.github/workflows/plaid-tests.yml
+++ b/.github/workflows/plaid-tests.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [windows-latest, macos-latest]
-        python-version: ["3.9", "3.12"]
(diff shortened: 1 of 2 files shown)


Samoed:
Yes, but tests were failed on 3.12 and 3.14 if I correctly remember

raphaelsty:
We will run the test on main for 3.10 and 3.14, I want the test to run fast-enough and not to consume too much bandwidth as python 3.9 is deprecrated in ST. Making the PR asap.

EDIT: Opened a PR to support python 3.14 but unfortunately there is an error with FastPlaid, as soon as tch-rs becomes compatible with Python 3.14 we will update PyLate: https://github.com/lightonai/pylate/pull/204

PR #190, opened 2026-01-06
[Draft] Bump ST

NohTow:
Just testing to bump ST (to extend NanoBEIR)


NohTow:
Tests seems fine out of the box, the modifications to NanoBEIR evaluator is not to fix it per say (it was still working fine) but to be able to specify another dataset_id to use another language

NohTow:
We should close this (because of #210), but I'm letting it open to remind me to update NanoBEIR

PR #195, opened 2026-02-18
Add ScaNN index backend with dtype-aware storage and tests

robro612:
## Summary
- add a new `ScaNN` index backend and export it from `pylate.indexes`
- support dtype-aware flattening/storage for fp16 and fp32, optional memory logging utilities, and scann optional dependencies in `pyproject.toml`
- add focused ScaNN unit coverage for dtype handling, verbosity normalization, embedding retrieval-by-docid, and error paths

## Test plan
- [x] `python -m compileall pylate/indexes/scann.py tests/test_scann.py`
- [x] `python -m pytest tests/test_scann.py -q`
- [x]  Test e2e with `python examples/evaluation/beir_dataset.py --index_type scann --dataset dataset_name nfcorpus`


raphaelsty:
Hi @robro612, this is a cool PR, LGTM, would it be possible to have a small benchmark on smallest dataset of BEIR benchmark like scifact and few small other to assert that the ndcg@10 is relevant ? As well as the QPS ?

robro612:
@raphaelsty certainly, I'll have those numbers as a consequence of testing the training PRs soon - keep in mind ScaNN is cpu-only so it's rather slow compared to FastPLAID, but was easier for me to use than Voyage even with the huge memory overhead of storing the flattened embs. I'll also take a look at the fixes the bot suggested ^.

robro612:
## Update: Copilot fixes + BEIR benchmarks

### Changes in this push
- Renamed `name` param to `index_name` to match convention used by PLAID/Voyager/FastPlaid
- Simplified redundant `.to(dtype=...)` in `add_documents`
- Added default value for `batch_size` in `add_documents` (no-op for ScaNN but required by `Base` interface)
- Fixed `retrieve.ColBERT` to properly route ScaNN through the candidate + rerank path (was incorrectly using the PLAID direct-return path)
- Updated retriever type annotations (`index: Base` instead of `Voyager | PLAID`)
- Added benchmark script (`examples/evaluation/benchmark_beir.py`)

### Benchmark results

Model: `lightonai/GTE-ModernColBERT-v1` (fp16), A100 GPU, k=10

#### nDCG@10

| Dataset | #Docs | #Queries | PLAID | ScaNN |
|---------|------:|--------:|------:|------:|
| nfcorpus | 3,633 | 323 | 0.3797 | 0.3796 |
| fiqa | 57,638 | 648 | 0.4520 | 0.4508 |
| scifact | 5,183 | 300 | 0.7622 | 0.7618 |
| trec-covid | 171,332 | 50 | 0.8391 | 0.8459 |

#### QPS (queries per second)

| Dataset | PLAID | ScaNN |
|---------|------:|------:|
| nfcorpus | 23.0 | 31.3 |
| fiqa | 21.4 | 6.7 |
| scifact | 21.2 | 7.1 |
| trec-covid | 16.2 | 3.4 |

#### Notes
- **ndcg@10 is virtually identical** between PLAID and ScaNN across all datasets (within 0.01)
- **QPS**: PLAID is faster on larger datasets due to GPU-accelerated retrieval; ScaNN (CPU-only) is faster on small datasets where the GPU overhead dominates
- ScaNN indexing is slower than PLAID on A100 (CPU-bound kmeans), but provides a simpler dependency story (pip-installable, no Rust/ColBERT compilation)
- ScaNN is a good option for users who want an easy-to-install approximate index without the PLAID toolchain

raphaelsty:
LGTM the MR is very clean @robro612, thank you for the evaluation results , I'll run the CI and then merge :)

examples/evaluation/beir_dataset.py
```diff
@@ -44,14 +46,22 @@
...
     args = parser.parse_args()
     dataset_name = args.dataset_name
+    index_type = args.index_type
     model_name = "lightonai/GTE-ModernColBERT-v1"
     model = models.ColBERT(
         model_name_or_path=model_name,
         document_length=300,
         query_length=query_len.get(dataset_name),
-    )
+    ).to(torch.float16)
```
NohTow: probably a nit because fp16 is almost == to fp32 but I wonder if this should be an option
should be noted that i am using fp16 for most of my benches these days to save some memory for large datasets (i need to fix bf16 models that outputs fp32 because of numpy and thus have to be recasted)

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+        if isinstance(verbose, bool):
+            return "init" if verbose else "none"
+        return verbose
+
+    def _log_retrieve(self) -> bool:
+        return self.verbose_level == "all"
+
+    def _build_searcher(self, embeddings: np.ndarray) -> None:
+        """Build the ScaNN searcher from embeddings (in-memory only)."""
+        build_start = time.time()
```
NohTow: Do we still need those though?
It's to run the bench right? I wonder if we should let bench params within final PR

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+            if self.verbose:
+                logger.info(
+                    f"[ScaNN] Building ScaNN searcher with {embeddings.shape[0]} vectors..."
+                )
+                logger.info(
+                    f"[ScaNN]   Parameters: num_leaves={self.num_leaves}, num_leaves_to_search={self.num_leaves_to_search}, training_sample_size={self.training_sample_size}, num_neighbors={self.num_neighbors}"
+                )
+
+        # Build ScaNN searcher
+        log_memory("Before scann.build()", self.verbose)
```
NohTow: A bit of a broad comment around the whole PR, but I wonder if we should leave time/memory profiling in the final merged thing

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+                )
+                if (
+                    self.num_leaves is not None
+                    or self.num_leaves_to_search is not None
+                    or self.training_sample_size is not None
+                ):
+                    logger.warning(
+                        f"[ScaNN]   WARNING: Manual parameters provided but will be ignored: num_leaves={self.num_leaves}, num_leaves_to_search={self.num_leaves_to_search}, training_sample_size={self.training_sample_size}"
+                    )
+        else:
```
NohTow, line 184: Sorry if dumb question but from my understanding here we were to use autopilot when params not set.
Seems like you are defining some defaults but not setting self.autopilot to True.
Were you referring to "auto tune" like use default or am I missing anything?

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+        """Load an existing index from disk. Raises an error if loading fails."""
+        index_path = self._get_index_path()
+        if index_path is None:
+            raise ValueError(
+                f"Cannot load index: index_folder or index_name not set. "
+                f"index_folder={self.index_folder}, index_name={self.index_name}"
+            )
+
+        metadata_path = index_path / "metadata.json"
+        doc_id_mapping_path = index_path / "doc_id_to_embedding_range.tsv"
```
NohTow: Am I a pain asking whether we could use the same type of processing than for the other indexes?
We went from sqlitdict to pickle, which should be pretty easy to implement (tsv -> dict pickled)

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+        try:
+            if self.verbose:
+                logger.info(f"[ScaNN] Loading existing index from {index_path}...")
+
+            # Load searcher - use absolute path to avoid path resolution issues
+            index_path_abs = index_path.resolve()
+            self.searcher = scann.scann_ops_pybind.load_searcher(str(index_path_abs))
+
+            # Load metadata (JSON)
+            with open(metadata_path, "r") as f:
```
NohTow, line 276: Very dumb question but I suppose you cannot get those from the scann object right?

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+                self.training_sample_size = metadata.get(
+                    "training_sample_size", self.training_sample_size
+                )
+                self.use_autopilot = metadata.get("use_autopilot", self.use_autopilot)
+                self.store_embeddings = metadata.get(
+                    "store_embeddings", self.store_embeddings
+                )
+
+            # Load doc_id_to_embedding_range (saved as TSV)
+            if doc_id_mapping_path.exists():
```
NohTow, line 296: See also this should simplify the logic here

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+                )
+
+            # Reconstruct position_to_doc_id from doc_id_to_embedding_range
+            if self.doc_id_to_embedding_range:
+                # Calculate total embeddings from the max end position
+                max_end = max(
+                    start + length
+                    for start, length in self.doc_id_to_embedding_range.values()
+                )
+                self.position_to_doc_id = np.empty(max_end, dtype=object)
```
NohTow, line 311: btw I do not mind having an array for this, it's different from the other but it's an optimization I discussed for FastPlaid and I especially do not mind because you are reconstructing it and not storing, it's neat

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+        batch_size: int = 128,
+    ) -> "ScaNN":
+        """Add documents to the index.
+
+        Note: This method only supports adding all documents at once.
+        Subsequent calls will raise an error.
+        batch_size is kept for API compatibility but not used.
+        """
+        # Enforce single add - check if documents already exist
+        if self._documents_added:
```
NohTow, line 431: RIP

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+            raise ValueError(
+                "ScaNN index only supports adding all documents at once. "
+                "Documents have already been added."
+            )
+
+        add_start = time.time()
+        if self.verbose:
+            logger.info(f"[ScaNN] Adding {len(documents_ids)} documents to index...")
+
+        log_memory("Start of add_documents", self.verbose)
```
NohTow: //

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+        # Get doc lengths and total count in one pass
+        doc_lengths = [emb.shape[0] for emb in documents_embeddings]
+        total_embeddings = sum(doc_lengths)
+        embedding_dim = documents_embeddings[0].shape[1]
+
+        # Preserve incoming dtype (fp16/fp32) to avoid holding two full flattened
+        # copies (e.g., one fp16 and one fp32) in memory at the same time.
+        first_emb = documents_embeddings[0]
+
+        def _to_np_dtype(dtype: object) -> type[np.float16] | type[np.float32] | None:
```
NohTow: No strong opinion but defining function in function has always been a bit odd to me, except for very specific use cases (avoid global import/lambda func)
Wdwt of putting it global/in a utility folder @raphaelsty?

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+        documents_ids
+            The documents IDs to remove.
+
+        Raises
+        ------
+        NotImplementedError
+            Document removal is not supported for ScaNN index.
+
+        """
+        raise NotImplementedError("Document removal is not supported for ScaNN index.")
```
NohTow, line 576: One possibility is to remove it from the list of ids
it's not perfect because it'll still be returned by the index (and thus not returning k elements), but just mentioning

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+
+    def __call__(
+        self,
+        queries_embeddings: list[list[int | float]],
+        k: int = 5,
```
NohTow: biggest nit of my life but default k for other indexes is 10

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+            Optional subset of document IDs to restrict search to.
+            Not yet implemented for ScaNN index.
+
+        Raises
+        ------
+        NotImplementedError
+            If subset is provided (not yet implemented).
+
+        """
+        if subset is not None:
```
NohTow: Wonder if we should expose the param at all then, it's not exposed for stanford plaid only fastplaid

pylate/indexes/scann.py
```diff
@@ -0,0 +1,814 @@
...
+                f"[ScaNN] Flattening {n_tokens_total} query tokens: {step_time:.4f}s"
+            )
+
+        # Query the index
+        step_start = time.time()
+        neighbors, distances = self.searcher.search_batched_parallel(
+            flattened_queries, final_num_neighbors=k
+        )
+        # replace NaN values with 0
+        if np.isnan(distances).any():
```
NohTow, line 608: can we add a small comment on why we are expecting nan? :o

pyproject.toml
```diff
@@ -42,6 +42,7 @@ dev = [
     "pytest-xdist >=3.6.0",
     "pytest-rerunfailures >= 15.0.0",
     "pytest >= 8.2.1",
+    "psutil >= 7.2.2",
```
NohTow: Why do we need it in main dep?

NohTow (commented):
Hey!
Thanks for the amazing work and sorry for the delay in the review, busy days!

I've added a few comments, most of them are nit but I figured it could help making things a bit cleaner (also some probable stupid questions, but I'd rather ask and find out I'm dumb rather than merging errors because I did not ask!)

Besides those, I think my main "comment" is that I wonder whether we should merge the part about time/memory profiling. It's very nice of you to have added all of those and go the extra miles for benching the things, but I wonder if it's something we expect in the merged indexes.
On a related note, I wonder if we should merge `examples/evaluation/benchmark_index_beir.py` and if so, I do not think it should be in this folder imho

tests/test_scann.py
```diff
@@ -0,0 +1,180 @@
+from __future__ import annotations
```
NohTow, line 1: Thanks for adding tests, btw!

pylate/indexes/scann.py
robro612: autopilot is a config setting directly in the ScaNN library. AFAICT it sets reasonable settings, but I don't know how exactly. Recalling my experiments, It's slower/more accurate than the defaults that I set which come directly from the XTR inference notebook.

robro612:
## Addressing review comments

Thanks for the thorough review @NohTow! Here's what was addressed:

| Comment | Resolution |
|---|---|
| Remove profiling/timing code | Removed all `time.time()` profiling, `log_memory()` calls, and `_log_retrieve()` method |
| Remove psutil dependency | Removed from both `scann` optional deps and `dev` deps |
| Remove benchmark script from PR | Removed (easy to recreate if needed from the BEIR boilerplate) |
| Use pickle instead of TSV for doc mapping | Switched `doc_id_to_embedding_range` serialization to pickle |
| Get metadata from ScaNN object directly | Investigated — `searcher.config()` has build params protobuf but not our app-level flags (`use_autopilot`, `store_embeddings`), so JSON sidecar is still needed / desirable IMO for clear visibility without load. |
| Autopilot clarification | `use_autopilot` is a ScaNN config flag; when True it calls `.autopilot()` and ignores manual params — handled correctly |
| Move nested function to utils | Extracted `_to_np_dtype` → `np_dtype_for()` in `pylate/indexes/utils.py` |
| `remove_document` filtering | Prefer not to patch behavior the underlying index doesn't offer |
| Default `k=10` | Fixed to match other indexes |
| `subset` param | Removed entirely (matches Voyager/StanfordPLAID which also don't expose it) |
| Add NaN comment | Added explanation that ScaNN can't always complete top-k |
| In-memory array for reconstructed embeddings | Kept as-is |

Also added a save/load round-trip test covering the pickle serialization.

pylate/indexes/scann.py
```diff
@@ -0,0 +1,681 @@
...
+                if (
+                    self.num_leaves is not None
+                    or self.num_leaves_to_search is not None
+                    or self.training_sample_size is not None
+                ):
+                    logger.warning(
+                        "[ScaNN]   autopilot() overrides manual configuration (num_leaves, num_leaves_to_search, training_sample_size)"
+                    )
+        else:
+            # Auto-tune parameters if not set
```
NohTow:
```suggestion
            # Use default if not set
```

NohTow:
Thanks for the PR!

PR #196, opened 2026-02-18
Make ColBERT prefix tokens optional for prefix-free models

robro612:
## Summary
- Wire the existing `add_special_tokens` constructor parameter to actually control prefix token insertion during tokenization
- When `add_special_tokens=False`, prefix tokens (`[Q] ` / `[D] `) are not prepended and `max_seq_length` is not reduced
- Default behavior (`add_special_tokens=True`) is unchanged — `[Q] ` / `[D] ` prefixes are added as before
- `add_special_tokens` is persisted in `config_sentence_transformers.json` for save/load round-trips
- Legacy models without the key in config default to `True`

This enables loading models (e.g. XTR) that don't use ColBERT-style marker tokens without incorrect tokenization.

## Test plan
- [x] Add test suite (`tests/test_add_special_tokens.py`) covering:
  - Init stores the flag correctly (default True, explicit True/False)
  - Prefix token insertion/skipping in `tokenize()` for queries and documents
  - `max_seq_length` adjustment (reduced by 1 only when prefixes are used)
  - Save/load round-trip preserves the setting
  - Legacy config without the key defaults to True


pylate/models/colbert.py
```diff
@@ -364,18 +366,12 @@ def __init__(
         self.is_hpu_graph_enabled = False
         # Override the configuration values with the provided arguments, if any. If not set and values have not been read from configs, set to default values.
         self.query_prefix = (
-            query_prefix
-            if query_prefix is not None
-            else self.query_prefix
-            if self.query_prefix is not None
-            else "[Q] "
+            query_prefix if query_prefix is not None else self.query_prefix or "[Q] "
```
NohTow: I'm not sure if you just had the previous version or if it's intentional but I tried rebasing and it did not fix?
We now have stronger check because for example the empty string check returns false.
We can discuss if it was done on purpose, but since you are not using the "trick" of using empty string to disable prefixes, I'm not sure

pylate/models/colbert.py
```diff
@@ -364,18 +366,12 @@ def __init__(
...
         )
         self.document_prefix = (
             document_prefix
             if document_prefix is not None
-            else self.document_prefix
-            if self.document_prefix is not None
-            else "[D] "
+            else self.document_prefix or "[D] "
```
NohTow: same here

pylate/models/colbert.py
```diff
@@ -85,7 +85,9 @@ class ColBERT(SentenceTransformer):
     document_prefix
         Prefix to add to the documents.
     add_special_tokens
```
NohTow: I checked and it's indeed me who added this even if it wasn't in ST and isn't used like that (we can forward the tokenizer param in tokenizer_kwargs

So we can definitely use this param, but I think it would be best to rename it to not confuse ppl with the HF param. Probably "add_prefixes" would be a good name?
Also, I know we already discussed that back then, but what do we think about setting an empty string to not add? No strong opinion, cc @raphaelsty

pylate/models/colbert.py
```diff
@@ -85,7 +85,9 @@ class ColBERT(SentenceTransformer):
...
-        Add the prefix to the inputs.
+        Whether to prepend query/document prefix tokens during tokenization. Set to False for
+        models that don't use ColBERT-style marker tokens (e.g. XTR). If None, uses the value
```
NohTow: Thanks for making the description clearer!

pylate/models/colbert.py
```diff
@@ -459,6 +455,14 @@ def __init__(
         if not self.do_query_expansion:
             self.attend_to_expansion_tokens = False
 
+        self.add_special_tokens = (
+            add_special_tokens
+            if add_special_tokens is not None
+            else self.add_special_tokens
+            if hasattr(self, "add_special_tokens")
```
NohTow: Any reason not to follow the usual way?

```python
self.add_special_tokens = (
            add_special_tokens
            if add_special_tokens is not None
            else self.add_special_tokens
            if self.add_special_tokens is not None
            else True
        )
```

(with possible rename ofc)

pylate/models/colbert.py
```diff
@@ -1087,9 +1091,10 @@ def tokenize(
         """
         # Set max sequence length based on whether the input is a query or document
         max_length = self.query_length if is_query else self.document_length
+        use_prefix = self.add_special_tokens
```
NohTow: why defining a new variable for this?
i think it won't be needed after renaming the variable correctly

NohTow:
Looks pretty good to me
Main comment will be renaming the variable, I am dumb for adding it this way back then (and not even plugging it...), but it is definitely confusing (proof: I was confused o/).
Else I just wanted to raise the discussion about using a boolean rather than empty string again but I honestly do not have a strong feeling, so I'll let @raphaelsty decide what he prefers

Finally there is this question on why your tests/default definition are different from main/usual definition, else I believe we good to go.
In my opinion, we should go more and more towards deprecating prefixes and use prompts. I added the prefixes logic to match closely og colbert back then, but honestly it's a pain to maintain outside of PyLate (e.g, vLLM) and prompt would be exactly the same.

raphaelsty:
I'll give it a look tomorrow, the init part of colbert is always tricky, I'll need dedicated focus time

robro612:
Bump @raphaelsty

raphaelsty:
Stress tested the MR on various models, LGTM, default behaviour stays the same :) 

There are few comments from Antoine that need to be resolved before merging that's all

NohTow:
I think we're good to merge
Maybe using "" for empty prefix is idiomatic, but it's following usual norms in the ST ecosystem and somehow I like it

raphaelsty:
LGTM, just pushed a small commit to update a single test, merging as soon as tests pass

PR #203, opened 2026-02-25
Replace sqlitedict by pickle

NohTow:
This PR swap sqlitedict by pickle for managing the ids of documents.

Reason: sqlitedict has become a bottleneck in our evaluation pipelines, noticed that when running FastPlaid in either PyLate or directly and the diff was huge.
Replacing with pickle solve most of the overhead.
Tried thinking of alternatives:
image

So all in all, I think pickle is a very solid alternative.
While building, I realized there are some "improvements" we can make or not, whether you think it's important/needed:
1. Right now, we lost the atomic operation, which means if it crashed mid operation, the index will be corrupted. One solution is to write the files to a temp file then rename once it's done. That said, it uses twice the storage during write (although it's dict, so not that big). Also, writing should be fast-ish so the probability of a crash is pretty small. Finally, we need our other part of the index to be resillient, I know stanford nlp is probably not.

2. We are _loading_ the dict for every search. Right now, most of my setup are a single search setup (evaluation), but maybe it adds overhead for multiple search. We could add the possibility of keep the dicts in RAM, but I think it's a bad default as it consume quite some RAM for very large indexes (1-2GB).

3. Should be noted that, for fast-plaid, at some point, Claude came with a design to lower the memory usage even more. One of the dict can be swapped by a list, because the fast-plaid indexes are dense. The other has to stay a dict, and this cannot work for stanford-nlp. Also allows to get rid of the bisect when deleting.

### DIFF
--- a/pylate/indexes/fast_plaid.py
+++ b/pylate/indexes/fast_plaid.py
@@ -2,13 +2,13 @@ from __future__ import annotations
 
 import logging
 import os
+import pickle
 import shutil
 from bisect import bisect_left
 
 import numpy as np
 import torch
 from fast_plaid import search
-from sqlitedict import SqliteDict
 
 from ..rank import RerankResult
 from .base import Base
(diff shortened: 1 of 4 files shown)


raphaelsty:
Could you remove the dependancy if we do not use sqlitedict anymore ?

NohTow:
Made the changes for Voyager index as well and ditched the dependency altogether
Asked Claude to create a migration script if people wants to migrate their index ; did not test it but the code looks ok and it's essentially just a dict translation (there is an option to not remove old values so it's safe to test)


[migrate_sqlite_to_pickle.py](https://github.com/user-attachments/files/25548193/migrate_sqlite_to_pickle.py)

raphaelsty:
Amazing PR :D LGTM

PR #204, opened 2026-02-25
Update CI to Python 3.10/3.14, bump fast-plaid to 1.4.6.2110, extend FastPlaid API

raphaelsty:
### CI / packaging

- Test matrix bumped to Python 3.10 and 3.14 only (was 3.10 / 3.12).
- `fast-plaid` bumped to `1.4.6.2110` (was `1.4.4`); FastPlaid index gains a `low_memory` parameter.
- `xtr-warp-rs` bumped to `2.0.3.*`. The `python_version < '3.14'` marker is dropped now that cp314 wheels exist on PyPI (via [pau-mensa/xtr-warp-rs#16](https://github.com/pau-mensa/xtr-warp-rs/pull/16) + [#17](https://github.com/pau-mensa/xtr-warp-rs/pull/17)).
- `voyager` and `scann` extras are restricted to `python_version < '3.14'` — neither ships cp314 wheels yet (voyager 2.1.0 maxes at cp312, scann 1.4.2 at cp313).
- `conftest.py` skips `pylate/indexes/voyager.py` and `pylate/retrieve/xtr.py` doctests when the corresponding optional backend is not installed. `tests/test_retriever.py::test_voyager_index` gains a `pytest.importorskip("voyager")` guard.
- Doctest expectations cleaned up to drop Rust stdout lines that don't appear in the new fast-plaid release.

### FastPlaid wrapper symmetry with WARP

`FastPlaid` and `WARP` now offer the same end-to-end API surface (`add_documents` / `remove_documents` / `update_documents` / `__call__`):

- **`update_documents(documents_ids, documents_embeddings)`** added to `FastPlaid` (and exposed via `PLAID`). fast-plaid's backend has no per-passage update API, so the wrapper implements update as `remove_documents` → `add_documents`: document IDs are preserved (callers keep the same keys), underlying plaid IDs may shift — same external contract WARP gives.
- **`seed` and `num_threads`** added to `FastPlaid.__init__` (and plumbed through `PLAID`), mirroring WARP's parameter names. These are now forwarded to fast-plaid's `create` / `update` / `search` calls. Previously `update` was called with only `documents_embeddings`, so `kmeans_niters` / `max_points_per_centroid` / `n_samples_kmeans` / `seed` / `use_triton_kmeans` silently fell back to backend defaults instead of the values the user configured at index construction.
- **`low_memory`** also exposed on the `PLAID` wrapper (already existed on `FastPlaid` but `PLAID` didn't forward it). Default `False`.

### Tests

- `test_fast_plaid_update_documents` mirroring `test_warp_update_documents` (add A/B/C → replace B's embedding with a space-themed one → assert B ranks for "rockets and space" and all three doc IDs still appear).

## Test plan

- [x] PyLate CI matrix passes on Python 3.10 and 3.14 (ubuntu / macos / windows where applicable).
- [x] Lint, documentation, and existing PLAID / fast-plaid / warp tests still pass.
- [ ] `test_fast_plaid_update_documents` runs in the ubuntu 3.10/3.14 `tests` jobs.

## Related

- [pau-mensa/xtr-warp-rs#16](https://github.com/pau-mensa/xtr-warp-rs/pull/16) — Python 3.14 support via PyO3 ABI3 forward compatibility (merged).
- [pau-mensa/xtr-warp-rs#17](https://github.com/pau-mensa/xtr-warp-rs/pull/17) — cibuildwheel bump + cp314 scoped to torch ≥2.9 (merged; cp314 wheels now on PyPI for `xtr-warp-rs == 2.0.3.{290,2100,2110}`).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,7 +9,7 @@ description = "A library for training and retrieval with ColBERT."
 readme = "README.md"
 authors = [{ name = "LightOn" }]
 license = "MIT"
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 classifiers = [
     "Programming Language :: Python :: 3",
     "Operating System :: OS Independent",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
     "ujson == 5.10.0",
     "ninja == 1.11.1.4",
     "fastkmeans == 0.5.0",
-    "fast-plaid>=1.2.4.260,<=1.3.0.290",
+    "fast-plaid>=1.4.4.270,<=1.4.4.290",
 ]
 keywords = []
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,12 +49,12 @@ dev = [
     "fastapi >= 0.114.1",
     "uvicorn >= 0.30.6",
     "batched >= 0.1.2",
-    "voyager >= 2.0.9",
+    "voyager >= 2.0.9; python_version < '3.14'",
     "typos >= 0.11.0",
 ]
 eval = ["ranx >= 0.3.16", "beir >= 2.0.0"]
 api = ["fastapi >= 0.114.1", "uvicorn >= 0.30.6", "batched >= 0.1.2"]
-voyager = ["voyager >= 2.0.9"]
+voyager = ["voyager >= 2.0.9; python_version < '3.14'"]
 scann = ["scann >= 1.4.2"]
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,7 +23,7 @@ dependencies = [
     "ujson == 5.10.0",
     "ninja == 1.11.1.4",
     "fastkmeans == 0.5.0",
-    "fast-plaid>=1.2.4.260,<=1.3.0.290",
+    "fast-plaid>=1.4.6.270,<=1.4.6.2110",
 ]
 keywords = []
 
@@ -49,14 +49,14 @@ dev = [
     "fastapi >= 0.114.1",
     "uvicorn >= 0.30.6",
     "batched >= 0.1.2",
-    "voyager >= 2.0.9",
+    "voyager >= 2.0.9; python_version < '3.14'",
     "typos >= 0.11.0",
 ]
 eval = ["ranx >= 0.3.16", "beir >= 2.0.0"]
 api = ["fastapi >= 0.114.1", "uvicorn >= 0.30.6", "batched >= 0.1.2"]
-voyager = ["voyager >= 2.0.9"]
-scann = ["scann >= 1.4.2"]
-warp = ["xtr-warp-rs == 2.0.2.*"]
+voyager = ["voyager >= 2.0.9; python_version < '3.14'"]
+scann = ["scann >= 1.4.2; python_version < '3.14'"]
+warp = ["xtr-warp-rs == 2.0.3.*"]
 
 
 [dependency-groups]
--- a/.github/workflows/plaid-tests.yml
+++ b/.github/workflows/plaid-tests.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [windows-latest, macos-latest]
-        python-version: ["3.9", "3.12"]
+        python-version: ["3.10", "3.14"]
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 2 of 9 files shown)


raphaelsty:
Investigating how to make fast-plaid compatible with python 3.14 version

https://github.com/lightonai/fast-plaid/pull/37

Blocker encountered with tch-rs, it's ok for now to not support python 3.14, we will as soon as tch-rs support recent PyO3 version

https://github.com/LaurentMazare/tch-rs/issues/1002

mchar7:
I see you hit the tch-rs wall for 3.14 support. I've been poking at the same issue.

I managed to get fast-plaid built and running on 3.14 with the ABI3 compatibility flag. Built a wheel locally and ran the PyLate and fast-plaid test suites and all pass.

This would just need to be added to `CIBW_ENVIRONMENT`:
```
PYO3_USE_ABI3_FORWARD_COMPATIBILITY=1
```

That would be a no-op for <3.14 and your existing `--exclude libtorch*` auditwheel flags handle the bundling correctly.

I'd get it if if you'd prefer to wait for tch-rs to properly support PyO3 0.28, but if you want to unblock 3.14 now I can throw together a PR for fast-plaid#37.

raphaelsty:
You rock @mchar7, would love to get a PR with the CIBW flag, thank you for the feedback

mchar7:
I pushed a few commits on top of your branch to get the 3.14 tests green.

- Bumped sentence-transformers and transformers to the 5.x line. This fixes the main issue where ModernBERT's @torch.compile decorators caused an import crash on 3.14. Also bumped ujson for cp314 wheels and synced the ruff target version to py310. (b820113)
- Transformers 5.x dropped overwrite_output_dir, so I removed it from the training tests and docs. Also skipped Voyager in test_retriever.py on 3.14 since there's no wheel yet. (460f973)
- Added lightonai/GTE-ModernColBERT-v1 to the parametrized load tests to catch this specific regression going forward. (bc70208)
- Fixed the PLAID tests on Windows 3.14. zlib-state (pulled in by ranx) doesn't have cp314 wheels and fails to build without zlib.h. Swapped --extra dev for --with pytest --with pytest-cov in the workflow since eval deps aren't needed there anyway. (4a3658c)

All workflows passing: CI run (https://github.com/mchar7/pylate/pull/1). Branch is at mchar7/pylate:python314-fix (https://github.com/mchar7/pylate/tree/python314-fix) if you want to scan through.

moritzwilksch:
What is the reason for the oddly tight `fast-plaid` pin both on main at the moment but also in this PR?

NohTow:
> What is the reason for the oddly tight fast-plaid pin both on main at the moment but also in this PR?

One reason is because we are using version to also align ship the version of cuda. Honestly at some point I wanted us to do something cleaner in the style of FA, that use a custom build file that tries to fetch precompiled wheels, but we never got the time.
But I think even for this purpose, we should still be able to lift a bit the range, the thing is that we are responsible for both packages and the updates should always be enhancements over previous versions (the package is still very young and a lot of progress are made regularly). I do not think we changed dependencies within fast-plaid, so any reason you would want it to be broader?

(One thing to note though, is that this time Raph added a new parameter which I do not want the default of fast-plaid to be the default of PyLate, and I think we won't have choice but to force the new version in PyLate as well if we add the param definition in PyLate)

moritzwilksch:
I see, thanks! I’m asking because I’m trying to build and package the latest pylate release on conda-forge, and it now looks like that will require building an older fast-plaid version first. Not a big deal at all, I just wanted to confirm that the pin is actually necessary.

More broadly, pylate feels like it’s becoming a foundational library for the IR community, kind of like sentence-transformers, so ideally the packaging around it stays as clean and painless as possible.

Anyway, I’m a huge fan of your work, and I can’t wait to try the new LateOn/DenseOn models once we get the infra side sorted :)

pyproject.toml
```diff
@@ -49,12 +49,12 @@ dev = [
     "fastapi >= 0.114.1",
     "uvicorn >= 0.30.6",
     "batched >= 0.1.2",
-    "voyager >= 2.0.9",
+    "voyager >= 2.0.9; python_version < '3.14'",
```
Samoed, line 52: Previously voyageer supported up to py3.12, but it seems that bumped python support https://github.com/spotify/voyager/pull/136, but they didn't push new release

PR #205, opened 2026-02-25
Fix faulty defaults

### LINKED ISSUE #192: Unexpected embedding warning when loading colbert-ir/colbertv2.0
Hi,

When I load the model colbert-ir/colbertv2.0 using pylate, I see the following warning:

> The new embeddings will be initialized from a multivariate normal distribution that has old embeddings' mean and covariance. As described in this article: https://nlp.stanford.edu/~johnhew/vocab-expansion.html. To disable this, use `mean_resizing=False`

The following is an example:
```
>>> from pylate import indexes, models, retrieve
>>> model = models.ColBERT(model_name_or_path="colbert-ir/colbertv2.0")
No sentence-transformers model found with name colbert-ir/colbertv2.0.
The new embeddings will be initialized from a multivariate normal distribution that has old embeddings' mean and covariance. As described in this article: https://nlp.stanford.edu/~johnhew/vocab-expansion.html. To disable this, use `mean_resizing=False`
```

Is this expected behaviour?
Will this affect the model’s retrieval performance?

Best regards,
Chuan Meng

NohTow:
Our tests to use default values were based on boolean test (if XX, else default)
Problem is, in Python, if [] returns False, same for int or empty string.
I already had this issue that I fixed for some of the params back then, but it happened again for skiplist.
This PRs explicitely test for None before defaulting, also change for stanford-nlp reading/defaulting.

I take the opportunity to also fix #192 by, if the stanford-nlp metadata _exists but no the query/doc prefixes_ (which is an edge case as it seems only colbert-v2 does not define them), then we default to unused0/1 (default of stanford-nlp), instead of nothing that then default to [Q] [D]. I fix it's valid because this is the default of stanford, and anyone who did not use those should have them in the metadata.

### DIFF
--- a/pylate/models/colbert.py
+++ b/pylate/models/colbert.py
@@ -284,18 +284,17 @@ class ColBERT(SentenceTransformer):
                         metadata = json.load(f)
                         # If the user do not override the values, read from config file
                         meta_query_token_id = metadata.get("query_token_id", None)
-                        if (
-                            self.query_prefix is None
-                            and meta_query_token_id is not None
-                        ):
-                            self.query_prefix = meta_query_token_id
-
+                        if self.query_prefix is None:
+                            if meta_query_token_id is not None:
+                                self.query_prefix = meta_query_token_id
+                            else:
+                                self.query_prefix = "[unused0]"
                         meta_doc_token_id = metadata.get("doc_token_id", None)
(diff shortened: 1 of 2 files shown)


raphaelsty:
LGTM

Issue #144, opened 2025-08-31
Dependency issue with ModernBERT Training


krypticmouse:
Hi Folks

I've been trying to train ColBERT on `ModernBERT-base` model with the following environment dependency:

```
Name: sentence-transformers
Version: 4.0.2

Name: transformers
Version: 4.50.0
```

I keep getting the following error on the training script provided in the docs: 
```
TypeError: ModernBertModel.__init__() got an unexpected keyword argument 'add_pooling_layer'
```

I tried `gte-modernbert-base` as well and got:

```
FileNotFoundError: [Errno 2] No such file or directory: 'Alibaba-NLP/gte-modernbert-base/sentence_xlnet_config.json'
```

```
FileNotFoundError: [Errno 2] No such file or directory: 'nomic-ai/modernbert-embed-base/sentence_xlnet_config.json'
```

Any fixes/alternate models/revision for this?

NohTow:
Hey,
Can you give your PyLate version?
Seems like an issue we fixed a looooong time ago
See https://github.com/lightonai/pylate/pull/52

krypticmouse:
Found the issue, apparently uv was reverting to 1.0.0 for pylate to resolve deps in env, after some adjustments and installing the latest version it worked out 👍 

Thank you!

NohTow:
Could you please add some information about what you did?
It might affect other users of uv as well and we are in the process of going towards using uv for the repo!

PR #210, opened 2026-03-24
feat: relax sentence-transformers and transformers version pins for 5.x compat

j-sperling:
## Summary

- Bump `sentence-transformers` from `== 5.1.1` to `== 5.3.0`
- Cap `transformers` at `< 5.0.0` (was `<= 4.56.2`) to allow 4.57.x while deferring v5 until numerical equivalence is confirmed
- Remove deprecated `overwrite_output_dir` from 3 test files (no-op on v4, required for eventual v5 support)

## Numerical verification

ColBERT embeddings are **bit-identical** between transformers 4.56.2 and 5.4.0 (both with ST 5.3.0, torch 2.9.0):

| Model | max abs diff | min cosine sim | Verdict |
|-------|-------------|----------------|---------|
| lightonai/GTE-ModernColBERT-v1 | 0.00 | 1.000000 | Bit-identical |
| lightonai/colbertv2.0 | 0.00 | 1.000000 | Bit-identical |

The known transformers v5 divergences (huggingface/transformers#42889, huggingface/transformers#43697) affect T5/encoder-decoder and vision models due to weight tying refactoring. BERT-based architectures used by pylate's ColBERT models are unaffected.

## Test plan

Tested with ST 5.3.0 + transformers 4.57.6 on Python 3.12 (Apple Silicon):

- [x] Model loading works: GTE-ModernColBERT-v1, ColBERT-Zero, colbertv2.0
- [x] Encoding produces correct shapes (per-token 128d embeddings)
- [x] 13/16 tests pass (3 training tests fail on MPS -- pre-existing, unrelated)
- [x] 1 model-loading test fails (jina-colbert-v2 HF config issue -- pre-existing)

## Motivation

The current exact pin on `sentence-transformers==5.1.1` prevents PyLate from being installed alongside other packages that depend on newer ST versions. This is a common integration pain point (see #144, #190).

The `transformers<=4.56.2` upper bound blocks 4.57.x which includes bug fixes and performance improvements.

## Related

- #190 (Draft: Bump ST to 5.2.0)
- #204 (CI updates for Python 3.10-3.14)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,17 +9,17 @@ description = "A library for training and retrieval with ColBERT."
 readme = "README.md"
 authors = [{ name = "LightOn" }]
 license = "MIT"
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 classifiers = [
     "Programming Language :: Python :: 3",
     "Operating System :: OS Independent",
 ]
 dependencies = [
-    "sentence-transformers == 5.1.1",
+    "sentence-transformers == 5.3.0",
     "datasets >= 2.20.0",
     "accelerate >= 0.31.0",
     "pandas >= 2.2.1",
-    "transformers >= 4.41.0, <= 4.56.2",
+    "transformers >= 4.41.0, <= 5.3.0",
     "ujson == 5.10.0",
     "ninja == 1.11.1.4",
     "fastkmeans == 0.5.0",
@@ -76,7 +76,7 @@ pylate = ["hf_hub/model_card_template.md", "py.typed"]
 [tool.ruff]
 line-length = 88
 indent-width = 4
-target-version = "py39"
+target-version = "py310"
 
 [tool.ruff.lint]
 select = ["E4", "E7", "E9", "F", "I", "D417"]
--- a/tests/test_kd.py
+++ b/tests/test_kd.py
@@ -42,7 +42,6 @@ def test_kd_training() -> None:
 
     args = SentenceTransformerTrainingArguments(
         output_dir="tests/kd",
-        overwrite_output_dir=True,
         num_train_epochs=1,
         per_device_train_batch_size=1,
         fp16=False,
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           enable-cache: true
           cache-dependency-glob: "**/pyproject.toml"
-          python-version: "3.9"
+          python-version: "3.10"
 
       - name: Run pre-commit
         run: |
--- a/.github/workflows/plaid-tests.yml
+++ b/.github/workflows/plaid-tests.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [windows-latest, macos-latest]
-        python-version: ["3.9", "3.12"]
+        python-version: ["3.10", "3.12"]
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -19,7 +19,7 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest]
-        python-version: ["3.9", "3.12"]
+        python-version: ["3.10", "3.12"]
         test-target: ["tests", "pylate"]
 
     steps:
--- a/tests/test_cached_contrastive.py
+++ b/tests/test_cached_contrastive.py
@@ -39,7 +39,6 @@ def test_contrastive_training() -> None:
 
     args = SentenceTransformerTrainingArguments(
         output_dir="tests/cached_contrastive",
-        overwrite_output_dir=True,
         num_train_epochs=1,
         per_device_train_batch_size=2,
         per_device_eval_batch_size=1,
--- a/tests/test_contrastive.py
+++ b/tests/test_contrastive.py
@@ -42,7 +42,6 @@ def test_contrastive_training() -> None:
 
     args = SentenceTransformerTrainingArguments(
         output_dir="tests/contrastive",
-        overwrite_output_dir=True,
         num_train_epochs=1,
         per_device_train_batch_size=1,
         per_device_eval_batch_size=1,


NohTow:
Thanks for the PR!
Let me run the test and try to have a look ASAP

Edit:
I might take a bit of time to review because:
1. I am quite drowning lately but most importantly
2. This is a major update for transformers, so I need to check with @tomaarsen if there isn't any pitfall 

In any case, we'll have to pin the exact version of ST to the latest one: we are not using `>=` because it can break when Tom release a new version. I know it's a pain and we also have to make updates manually each time, but this is the only way to keep something stable. This is the main argument in favor of merging PyLate upstream in ST.

pyproject.toml
```diff
@@ -15,11 +15,11 @@ classifiers = [
     "Operating System :: OS Independent",
 ]
 dependencies = [
-    "sentence-transformers == 5.1.1",
+    "sentence-transformers >= 5.1.1",
```
Samoed:
```suggestion
    "sentence-transformers >= 5.1.1, <6.0.0",
```

matospiso:
if you're using [uv](https://docs.astral.sh/uv/), you can temporarily avoid this inconvenience by adding

```
[tool.uv]
override-dependencies = [
    "transformers>=some_version",
    "sentence-transformers>=another_version",
]
```

to your `pyproject.toml` (until this PR is merged)

NohTow:
Hey,
I chatted quickly with @tomaarsen and the new ST version indeed allows to use Transformers v5 but I have a few things to consider before making this update:
1. Transformers v5 seems to yield different values than v4.X for some models. Until we know why (Tom is currently working on multimodal update and did not have time to dig further), I would be a bit worried about lifting the version lock, as any new install would use v5 by default
2. Although all tests are passing, I need to take a bit of time to check the release note, as Tom did some internal modifications and I want to make sure everything is fine


I am sorry to be slow with some updates lately, I've been quite busy, but as raised above, you can change overcome those issues locally. Is there any very blocking setup that you are encountering? I'll try to fix everything ASAP

j-sperling:
Thanks for the detailed feedback @NohTow -- totally understand both concerns. I dug into the transformers v5 numerical issue before responding.

## ColBERT numerical comparison: transformers 4.56.2 vs 5.4.0

I ran pylate's `encode()` on identical inputs with transformers 4.56.2 and 5.4.0 (both with ST 5.3.0, torch 2.9.0):

| Model | max abs diff | min cosine sim | Verdict |
|-------|-------------|----------------|---------|
| `lightonai/GTE-ModernColBERT-v1` | 0.00 | 1.000000 | Bit-identical |
| `lightonai/colbertv2.0` | 0.00 | 1.000000 | Bit-identical |

The known transformers v5 divergences (huggingface/transformers#42889 for T5 weight tying, huggingface/transformers#43697 for RTDetrV2) affect encoder-decoder and vision model families. BERT-based architectures -- which all pylate ColBERT models use -- produce identical outputs.

Happy to test additional models if you'd like.

## Updated PR

Based on your feedback, I've updated the branch:

1. **ST pin**: `== 5.3.0` (exact pin per your policy)
2. **Transformers**: removed the `<= 4.56.2` ceiling, kept the original `>= 4.41.0` floor unchanged
3. **`overwrite_output_dir` removal**: kept (needed for transformers 5.x `TrainingArguments`)

If you'd prefer to keep the transformers ceiling until Tom confirms independently, I can revert that part and just ship the ST bump. Your call.

pyproject.toml
```diff
@@ -15,11 +15,11 @@ classifiers = [
     "Operating System :: OS Independent",
 ]
 dependencies = [
-    "sentence-transformers == 5.1.1",
+    "sentence-transformers == 5.3.0",
     "datasets >= 2.20.0",
     "accelerate >= 0.31.0",
     "pandas >= 2.2.1",
-    "transformers >= 4.41.0, <= 4.56.2",
```
TheAdamEvans, line 22: The transition to 5.*+ is a breaking change and feels too big right now.

In this PR, perhaps we can compromise and make this change to:
"transformers >= 4.41.0, <= 4.57.1",

.gitignore
```diff
@@ -166,3 +166,8 @@ uv.lock
 indexes/*
 pylate-index/*
 /docs/api/
+
+# Agent config (generated/local)
+.memsearch/
+AGENTS.md
+CLAUDE.md
```
TheAdamEvans: Don't need this addition in this PR.

AGENTS.md is another whole discussion and task.

TheAdamEvans (changes requested):

pyproject.toml
TheAdamEvans: Should be no breaking changes this way --> `transformers <= 4.57.1`

j-sperling:
Thanks @TheAdamEvans. Reverted the `.gitignore` additions -- clearly out of scope.

For the transformers ceiling: updated to `<= 4.57.1` as you suggested. I'd initially gone with `< 5.0.0` to avoid the pin going stale on each patch, but a tight ceiling is the safer call while the v5 numerical story is still being sorted out (see discussion above with @NohTow).

Kept the `overwrite_output_dir` removal -- it's a no-op on v4 (default was already `False`) and will be needed whenever v5 support lands.

NohTow:
Hey,
Sorry for the delay, Raphael and I were at ECIR last week and are in vacation this week
Will focus on the various PRs when I get back
Apologies for the delay again

tomaarsen:
Hello!

I've ran some more tests between `transformers` v4 and v5, and the discrepancies I spotted earlier were likely caused by differences in default options for `dtype` and `attn_implementation`. With those fixed, I'm only getting some discrepancies for ~3 rerankers and a handful of Sparse Encoder models, and no regressions for dense embedding models, across a total of 169 model tests.
In short, moving to transformers v5 seems pretty safe.

- Tom Aarsen

TheAdamEvans:
No worries team, I know you have a lot on and these are small things, am patient! 😄 Looking to get PyLate included in Canva's monorepo. It can be tricky because the entire company has to share the exact same version specs. Thanks @j-sperling for updating 🙏 

Re: sentence-transformers, I'd selfishly prefer <=5.1.2 because that's what we have here today and will simplify things for me later on the next PyLate release.

@j-sperling could I convince you to also update `ninja` to be widened to >=1.11.1? That === pin conflicts with other packages that currently resolve to `1.11.1.1` (a more commonly available version from what I can see).

Happy to open separate issues for these if we'd prefer.

pyproject.toml
TheAdamEvans, line 26: Might I also suggest
```suggestion
    "fast-plaid>=1.2.4.260,<=1.4.6.280",
```

j-sperling:
Thanks for testing, @tomaarsen -- 169 models is thorough. That's load-bearing evidence for reconsidering the transformers ceiling.

@TheAdamEvans — on the ST pin: NohTow set an explicit policy Mar 24 that we pin ST to the exact latest version to keep it stable. The current branch (== 5.3.0) follows that. Downgrading to <= 5.1.2 for Canva's monorepo constraints conflicts with that policy. That decision lands with @NohTow when he's back.

On ninja and fast-plaid: neither changed in this PR (diff shows only ST + transformers lines touched). You offered to open separate issues for these — let's do that. Keeps this PR focused on ST 5.x compatibility and avoids scope creep.

Deferring to NohTow on the transformers ceiling with Tom's test data in the record.

j-sperling:
## Update: ST 5.4.1 + Python 3.10 floor

I've pushed a commit (4b98459) that updates this PR based on new information since Apr 6:

**sentence-transformers 5.4.1 (released today)**
- Newer than the PR's 5.3.0 pin.
- tomaarsen (ST maintainer) confirmed compatibility with transformers v5 across 169 models, including dense encoders.
- ST 5.4 explicitly advertises 'full compatibility with transformers v5'.
- ST 5.2+ requires Python >= 3.10, so updating the floor is necessary for correctness.

**Python floor now >= 3.10**
- Matches raphaelsty's PR #204 and the draft in NohTow's PR #190.
- Both maintainer PRs independently concluded 3.10 is the right floor.
- ST 5.2+ cannot run on 3.9, so this was required for the PR to be installable.

**On the remaining outstanding reviews:**

- **transformers ceiling** (@TheAdamEvans): The pinned `<= 4.57.1` is reasonable; I've deferred this to you, @NohTow, as the final call.
- **ninja + fast-plaid** (@TheAdamEvans): raphaelsty's PR #204 is already open and handles the fast-plaid range expansion. ninja is separately scoped and can be addressed in its own PR.

Thanks all for the review and context.

tests/test_kd.py
```diff
@@ -42,7 +42,6 @@ def test_kd_training() -> None:
 
     args = SentenceTransformerTrainingArguments(
         output_dir="tests/kd",
-        overwrite_output_dir=True,
```
TheAdamEvans, line 45: I think if we're sticking with `transformers <= 4.57.1` then these changes are no longer needed

TheAdamEvans (approved):

TheAdamEvans:
@j-sperling I think we're almost there! Thanks for hanging in here with me.

Am happy with the `sentence-transformers` pin, and will discuss the ninja + fast-plaid separately on #204 .

If we are not going to go to 5.0+ then those `overwrite_output_dir` changes are not necessary. Perhaps we remove those in this PR and then it will be ready for the maintainers approval. I see they are no-op though so whatever you and they prefer.

NohTow:
Hey all,
Thanks a lot for your work and deeply apologies for the delay, after the vacations we had to focus on [the release of DenseOn & LateOn](https://x.com/antoine_chaffin/status/2046609241918579019?s=20)
Now that's it done, I can update PyLate!

pyproject.toml
NohTow: I believe the main issue right now is that, fast-plaid update introduced a new low_memory parameter, that I want to be set to False by default in PyLate because it is a better default imho (it's much slower, so people should not use it, unless they _have_ to)
Also there is some issues with Python 3.14, so I would vote to leave out the update of fast-plaid to the [dedicated PR](https://github.com/lightonai/pylate/pull/204) with the discussions and stuff, but please feel free to drop in this one!

tests/test_kd.py
NohTow: I think it's ok since it was for tests anyways, this should be run on temp folders and never re-use an existing one I believe?

For context, `overwrite_output_dir` has been [deprecated](https://github.com/huggingface/transformers/pull/41323) and overwriting is now the default, unless you use `resume_from_checkpoint` and point towards the latest checkpoint.

NohTow: Not sure why you are saying those are no-op though @j-sperling
Default was false indeed, but the code was setting it to through, to avoid re-using anything from previous tests (although again, not sure it can be the case, but not an expert with Git jobs, cc @raphaelsty)

With the v5, having nothing is the same behavior has setting it to true prior to v5 so we're good for v5

pyproject.toml
```diff
@@ -9,17 +9,17 @@ description = "A library for training and retrieval with ColBERT."
...
 license = "MIT"
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 classifiers = [
     "Programming Language :: Python :: 3",
     "Operating System :: OS Independent",
 ]
 dependencies = [
-    "sentence-transformers == 5.1.1",
+    "sentence-transformers == 5.4.1",
```
NohTow:
```suggestion
    "sentence-transformers == 5.3.0",
```

NohTow: 5.3.0 version is the latest version of ST before the multimodal update
I'll handle the multimodal update on a separate PR, too many changes to handle here

pyproject.toml
```diff
@@ -9,17 +9,17 @@ description = "A library for training and retrieval with ColBERT."
...
     "datasets >= 2.20.0",
     "accelerate >= 0.31.0",
     "pandas >= 2.2.1",
-    "transformers >= 4.41.0, <= 4.56.2",
+    "transformers >= 4.41.0, <= 4.57.1",
```
NohTow:
```suggestion
    "transformers >= 4.41.0, <= 5.3.0",
```

NohTow: From @tomaarsen's comments/discussion, it seems that transformers 5.X is safe, as long as we update ST to a version that supports it. I am setting it to 5.3.0 because it's the latest version Tom tested with the ST 5.3.0 (yes, confusing naming lol)

NohTow (commented):
So, from what I understood, the reasons transformers v5 was "regressing" was due to:
1. `torch_dtype` defaulting to the values in the `config.json` instead of `float32`. This is fine because bf/fp16 are safe, just different. Also settable by the users anyways
2. The `attn_implementation` default to sdpa instead of eager. Here again, should be mostly fine, except if there are issue with the sdpa implementation, but that's not on us/can be set by the user (although I also learned that ModernBERT's default is no longer FA since 5.2. We should start setting it more explicitely imho)

The biggest bump for ST I am allowing for this PR is 5.3.0, as this is the latest version _before_ the multimodality update, which require a lot of change that I'll do in another PR. 
Since transformers 5.X seems to be safe-ish, we can lift the version to 5.3.0, which isn't the latest version per say, but the latest Tom tested with the ST 5.3.0.

I will be reading a bit the ST updates to make sure there's not anything missing not catched by our tests.

raphaelsty:
Hi, reviewing the changes, it's ok the CI fail for 3.9

tomaarsen:
ST dropped support for Python 3.9. Normally, that support is dropped when the security patches stop: https://endoflife.date/python

- Tom Aarsen

NohTow:
Yeah that's fine I pinged @raphaelsty on this matter
We bumped it in the pyproject but also need to bump it in the CI, as done in #204 
As we are only bumping the lower bound, we shouldn't have any issue with fast-plaid

Edit because I realized merging comments make comments disappear: 
- Bumping ST to 5.3.0 because it's the latest before multimodality, which we'll tackle in another PR.
- Bumping transformers to 5.3.0 because it's the latest version tested with ST 5.3.0. Also, it's fine to bump, earlier concern from Tom were because the default `torch_dtype` is read from the config instead of default to `float32`. We do not care because bf/fp16 are fine and it can be set by the user anyways. Also, default `attn_implementation` is now sdpa instead of eager. We do not care because it can be set by the user and sdpa should be fine, although it might be broken for some models (I also learned that since 5.2, ModernBERT default is no longer FA, so we should start defining it more explicitely)
- overwrite_output_dir is useless in tests right now anyways as we're deleting the folders manually
- fast-plaid updates will be handled in the [dedicated PR
](https://github.com/lightonai/pylate/pull/204)

tests/test_kd.py
NohTow: Actually, we're deleting the folder in the end of the tests, so yeah, it's fine

raphaelsty:
Amazing PR, I did update the CI and removed support for Python 3.9. I fixed few artefacts related to python 3.9. Then double check few things, LGTM, as soon as the CI pass I'll merge 😊🥳

PR #211, opened 2026-04-08
Add support for WARP

pau-mensa:
## Summary

- Add a new **WARP** index backend wrapping [xtr-warp-rs](https://github.com/pau-mensa/xtr-warp-rs) and export it from `pylate.indexes`
- Implement `add_documents`, `remove_documents` (with auto-compaction), and search with optional document ID filtering (shared or per-query)
- Add `update_documents` and `compact` as new public methods not present in the base Index class. `update_documents` replaces embeddings in-place without delete+add, `compact` physically removes tombstoned passages, empty centroids and reclaims disk space
- Expose all xtr-warp-rs hyperparameters via two config dataclasses (`WARPSearchConfig`, `WARPIndexingConfig`) to keep the constructor clean
- Extract shared embedding conversion utilities (`convert_embeddings_to_torch`) into `pylate/indexes/utils.py` for reuse across WARP and FastPLAID
- Add xtr-warp-rs as an optional dependency
- Add 12 focused WARP unit tests

## Tests

- `python -m pytest tests/test_warp.py`

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,6 +409,8 @@ Sample Output:
 ]
 ```
 
+Also note that a [WARP](https://github.com/pau-mensa/xtr-warp-rs) backend is also available via `indexes.WARP` (install with `pip install "pylate[warp]"`). WARP is an index to be used with [XTR-trained models](https://lightonai.github.io/pylate/documentation/training/#xtr-training) and might not work well with every model. It uses approximations that make retrieval faster and cheaper but can degrade performance on models that were not trained for them; see the [retrieval documentation](https://lightonai.github.io/pylate/documentation/retrieval/#xtr-retrieval) for the model compatibility notes and a concrete example.
+
 &nbsp;
 
 ## Reranking
--- a/docs/documentation/retrieval.md
+++ b/docs/documentation/retrieval.md
@@ -246,40 +246,131 @@ index = indexes.PLAID(
 )
 ```
 
+---
 
 ## XTR Retrieval
 
-PyLate's two-stage indexes (e.g. `retrieve.ColBERT` on a `ScaNN` or `Voyager` index) all share the same first stage: for each query token, the index returns its top-`k_token` document tokens with similarity scores. What differs is the **scoring step that turns those hits into a per-document score**. `retrieve.ColBERT` pools the candidate documents from the hits, fetches their embeddings from the index, and re-scores them with the full MaxSim. [XTR](https://arxiv.org/abs/2304.01982) (conteXtual Token Retrieval) instead **scores documents directly from the per-token-hit scores** themselves. For each query token, a document gets credit for the maximum score it received in that token's top-`k_token` list, summed across query tokens, with the **minimum retrieved score** imputed for query tokens that didn't surface any of its tokens (matching the original XTR paper).
+[XTR](https://arxiv.org/abs/2304.01982) (conteXtual Token Retrieval, Lee et al. NeurIPS 2023) is a multi-vector retrieval approach that changes both how the model is trained and how documents are scored. PyLate exposes XTR retrieval through the **WARP** backend, which packages it as a fast end-to-end IVF index via the [`xtr-warp-rs`](https://github.com/pau-mensa/xtr-warp-rs) Rust engine ([Jha et al., SIGIR 2025](https://arxiv.org/abs/2501.17788)). For best results, pair WARP with a model trained on the XTR objective via `pylate.scores.XTRScores`. See [XTR Training](training.md#xtr-training) for the training side.
 
-Swapping `retrieve.ColBERT` for `retrieve.XTR` on the same token index gives you XTR-style scoring with no re-embedding step, no full MaxSim rerank, and no need for the index to keep the original document embeddings around in `ScaNN`. Due to the coarser approximation of the scores, XTR extends its default token retrieval parameter `k_token` to `10_000`, greater than ColBERT's `100`:
+### How XTR scoring works
 
-```python
-from pylate import indexes, retrieve
+The classic ColBERT scoring step pools the candidate documents returned by a first-stage index, fetches their full embeddings, and re-scores them with the full MaxSim. [XTR](https://arxiv.org/abs/2605.00646) instead **scores documents directly from the per-token-hit scores** themselves: for each query token, a document gets credit for the maximum score it received in that token's top-`k_token` list, summed across query tokens, with the **minimum retrieved score** imputed for query tokens that did not surface any of its tokens (matching the original XTR paper). In PyLate, you can select between using ColBERT (full MaxSim scoring) or XTR (min score imputation) simply by using `retrieve.ColBERT` or `retrieve.XTR`.
 
-index = indexes.ScaNN(
-    index_folder="pylate-xtr-index",
-    index_name="my_documents",
-    override=True,
-    store_embeddings=False,  # XTR does not need cached document embeddings, ColBERT does.
-)
+WARP implements this scoring natively, with aggressive IVF-based approximations on top.
 
-# Encode and add documents to the index...
+### WARP (end-to-end)
 
-retriever = retrieve.XTR(index=index)
+WARP is the equivalent of PLAID for XTR retrieval, and `xtr-warp-rs` is an implementation of WARP based on FastPLAID. It is faster than FastPLAID because it uses more approximations. Latency and memory footprint at retrieval time go down, and retrieval quality stays competitive on standard IR benchmarks. Use WARP when search latency or RAM is your bottleneck.
 
-# Encode queries...
+#### Model compatibility
 
-results = retriever.retrieve(
-    queries_embeddings=queries_embeddings,
-    k=100,
-    k_token=10_000,
-)
+Unlike PLAID, WARP does not work equally well with every multi-vector model: its centroid-pruning approximations rely on a relatively flat token-score distribution, which standard ColBERT training does not always produce. When the model fits, WARP is both fast and accurate; when it does not, recall can drop noticeably.
 
-print(results)
+[Jha et al. (2026)](https://arxiv.org/abs/2605.00646) show that training with the [XTR](https://arxiv.org/abs/2304.01982) objective flattens the score distribution and recovers the model's retrieval quality on IVF-based engines, WARP included. PyLate supports XTR training natively as a drop-in replacement for ColBERT's score function. See [XTR Training](training.md#xtr-training) for how to wire `scores.XTRScores` (or `scores.XTRKDScores` for distillation) into your loss.
+
+In practice, you should benchmark both PLAID and WARP on your own data with your own model and pick whichever gives you the best speed/recall trade-off for your use case.
+
+#### Installation
+
+WARP is an optional dependency; install it explicitly:
+
+```bash
+pip install "pylate[warp]"
 ```
 
-???+ info "`store_embeddings=False`"
-    Because XTR scores documents straight from the index hits, the original document embeddings are never read again at retrieval time. Setting `store_embeddings=False` keeps the on-disk index much smaller — no need for `ScaNN` tostore the original document embeddings.
+Wheels are torch-version-suffixed (`xtr-warp-rs == 2.0.2.290` ships against torch 2.9.0, `2.0.2.280` against 2.8.0, etc.). The pylate extra resolves to the build that matches your installed torch.
+
+#### Indexing and retrieval
+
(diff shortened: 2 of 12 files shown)


pau-mensa:
## Notes

- WARP supports metadata filtering now, but I did not add it because I saw that the other indices don't have it
- WARP also supports streamed indexing, which can cut VRAM usage by 66%, but that would require changing things that seem a bit out of scope. Still, it is a powerful feature, let me know if you want me to add it

arn4:
I think the new version for `xtr-warp-rs` broke this pull request, I had to go back to v1.1.5 to pass the tests

pau-mensa:
I must've missed something when I did the release, let me check and I'll fix it

raphaelsty:
Hi @pau-mensa I'm doing a pass on the MR to merge it :)

raphaelsty:
The `WARPSearchConfig` / `WARPIndexingConfig` dataclasses look nice and the search-vs-indexing split is genuinely clean, but it's the only place in the codebase doing it (the other backends pass everything as flat kwargs on `__init__`). To stay consistent, I flattened the WARP constructor to match. We can revisit and roll out the dataclass approach across all backends in a dedicated MR if we want to standardize on that style.

NohTow:
I agree on the take of @raphaelsty (as discussed earlier).
This is genuinely but we would need to do it also for PLAID (and other indexes) and i'm still not sure how to that without breaking changes or creating a kwargs hell.

README.md
```diff
@@ -328,7 +328,7 @@ queries = Dataset.from_list(mapping=queries)
 
 ## Retrieval
 
-PyLate provides an efficient index with [FastPLAID](https://github.com/lightonai/fast-plaid). Simply load a ColBERT model and initialize the index to perform retrieval.
+PyLate provides an efficient index with [FastPLAID](https://github.com/lightonai/fast-plaid). Simply load a ColBERT model and initialize the index to perform retrieval. A second backend, [WARP](https://github.com/pau-mensa/xtr-warp-rs), is also available via `indexes.WARP` (install with `pip install "pylate[warp]"`). See the [retrieval documentation](https://lightonai.github.io/pylate/documentation/retrieval/#warp) for a concrete example.
```
NohTow: We should make it clear that it isn't compatible with all models without degradation, see documentation

raphaelsty:
Small follow-up: dropped the public `compact()` method and the `compact_threshold` constructor knob. `remove_documents` now always triggers a full compaction inline, so tombstoned passages and empty centroids are reclaimed on every delete with no extra step from the caller. Same rationale as the dataclass flatten: keeps the retrieval pipeline surface uniform across `PLAID` and `WARP` so users can swap one for the other without learning new lifecycle calls.

Also added `tests/test_delete_parity.py` — parametrized over both backends — pinning that deleting a "lower" id keeps higher ids addressable, and that re-adding a previously deleted id works. FastPlaid renumbers internally, WARP tombstones; the user-facing id layer hides both, and the test makes that contract explicit.

README.md
raphaelsty: done :)

docs/documentation/retrieval.md
```diff
@@ -246,6 +246,94 @@ index = indexes.PLAID(
...
+    batch_size=32,
+    is_query=True,
+)
+scores = retriever.retrieve(queries_embeddings=queries_embeddings, k=10)
+print(scores)
+```
+
+### Tuning
+
+All search and indexing hyperparameters are flat keyword arguments on `indexes.WARP(...)`, the same constructor style as `indexes.PLAID`. `n_ivf_probe` shares its name and meaning with the PLAID parameter. The other WARP-specific search knobs (`bound`, `t_prime`, `max_candidates`, `centroid_score_threshold`) default to `None`, which lets `xtr-warp-rs` auto-tune them from the index. Leave them alone unless you have a reason. Indexing params (`nbits`, `kmeans_niters`, `max_points_per_centroid`, ...) follow the PLAID conventions.
```
robro612: I want to reiterate my worries about the `xtr-warp-rs` parameters being undertuned for retrieval effectiveness. If we default our constructor parameters to None and let it's autotune take over, I think users will see a disappointingly low top-line nDCG/Recall number and drop WARP, even if it is blazing fast. 

@pau-mensa would you object to a safer default hyperparams (just set in the pylate WARP constructor) of e.g. nprobe=32, t'=100k with a louder callout that this is a _conservative but still faster than FastPlaid_ config that users should tinker with?

docs/documentation/retrieval.md
```diff
@@ -246,6 +246,94 @@ index = indexes.PLAID(
...
+
+WARP is faster than FastPLAID because it approximates more aggressively. It prunes the centroid space per query token and uses an error-aware merge to skip work that does not change the final ranking. Latency and memory footprint at retrieval time go down, and retrieval quality stays competitive on standard IR benchmarks. Use WARP when search latency or RAM is your bottleneck.
+
+### Model compatibility
+
+Unlike PLAID, WARP does not work equally well with every multi-vector model: its centroid-pruning approximations rely on a relatively flat token-score distribution, which standard ColBERT training does not always produce. When the model fits, WARP is both fast and accurate; when it does not, recall can drop noticeably.
+
+[Jha et al. (2026)](https://arxiv.org/abs/2605.00646) show that training with the [XTR](https://arxiv.org/abs/2304.01982) objective flattens the score distribution and recovers WARP's quality on top of any IVF-based engine. PyLate supports XTR training natively as a drop-in replacement for ColBERT's score function — see [XTR Training](training.md#xtr-training) for how to wire `scores.XTRScores` (or `scores.XTRKDScores` for distillation) into your loss.
```
robro612: > recovers WARP's quality on top of any IVF-based engine

this doesn't exactly type-check to me: WARP is an instance of an IVF-based engine. Should it read "recovers the model's quality...", "recovers the model's quality on any IVF-based engine", or "recovers WARP's quality."

docs/documentation/retrieval.md
pau-mensa: I have no problems with it and it actually goes with the PR's flavor of fitting WARP into the same interface the other indices expose.

However, as a practitioner and someone that runs this in production systems today I'd never trade the 5-10x performance gains that the auto-tuned parameters give you for the ~2% (relative) improvement on retrieval metrics that bumping up `nprobe` and `t_prime` offer.

Maybe the compromise here is to set these hyperparams (`nprobe=32`, `t_prime=100_000`) but give more clarity in the docs as to what the trade offs are? @NohTow @raphaelsty do you have an opinion here?

NohTow: I really have no strong opinion
I said to @robro612 that I prefer to have slower defaults so the user never wonders "wow ok but are those results bad because of the index or my model?", but in the case of XTR, I can understand that the goal is speed so it would make sense as well. I am very fine with going with slower defaults and adding a lot of info in the documentation, but I can understand if you think default should be more aggressive as well

pau-mensa: Alright, added the defaults @robro612 mentioned and an extra explanation in the retrieval docs. You can check the diff [here](https://github.com/lightonai/pylate/compare/460c7509423c7d769b5d0fb6c96912ee07f961ac..0fd12f75692a4487e11c69e348ec0a5083aa9604)

I think this fits nicely with the PR's flavor! I will say though that having to explicitly set `nprobe` and `t_prime` to `None` in order to boost the latency (with the quality tradeoff) seems pretty unintuitive, but is difficult to improve it unless the way pylate interacts with indices change, out of scope right now

docs/documentation/retrieval.md
```diff
@@ -246,40 +246,129 @@ index = indexes.PLAID(
...
+
+### How XTR scoring works
+
+The classic ColBERT scoring step pools the candidate documents returned by a first-stage index, fetches their full embeddings, and re-scores them with the full MaxSim. XTR instead **scores documents directly from the per-token-hit scores** themselves: for each query token, a document gets credit for the maximum score it received in that token's top-`k_token` list, summed across query tokens, with the **minimum retrieved score** imputed for query tokens that did not surface any of its tokens (matching the original XTR paper). WARP implements this scoring natively, with aggressive IVF-based pruning that skips work which would not change the final ranking.
+
+### WARP (end-to-end)
+
+WARP is faster than FastPLAID because it approximates more aggressively. It prunes the centroid space per query token and uses an error-aware merge to skip work that does not change the final ranking. Latency and memory footprint at retrieval time go down, and retrieval quality stays competitive on standard IR benchmarks. Use WARP when search latency or RAM is your bottleneck.
```
NohTow: WARP is the equivalent of PLAID for XTR retrieval. XTR-WARP is faster than...

NohTow:
```suggestion
WARP is the equivalent of PLAID for XTR retrieval and xtr-warp-rs is an implementation of WARP based on FastPLAID. xtr-warp-rs is faster than FastPLAID because it uses more approximations. Latency and memory footprint at retrieval time go down, and retrieval quality stays competitive on standard IR benchmarks. Use WARP when search latency or RAM is your bottleneck.
```

docs/documentation/retrieval.md
```diff
@@ -246,40 +246,129 @@ index = indexes.PLAID(
...
+
+### How XTR scoring works
+
+The classic ColBERT scoring step pools the candidate documents returned by a first-stage index, fetches their full embeddings, and re-scores them with the full MaxSim. XTR instead **scores documents directly from the per-token-hit scores** themselves: for each query token, a document gets credit for the maximum score it received in that token's top-`k_token` list, summed across query tokens, with the **minimum retrieved score** imputed for query tokens that did not surface any of its tokens (matching the original XTR paper). WARP implements this scoring natively, with aggressive IVF-based pruning that skips work which would not change the final ranking.
```
NohTow:
```suggestion
The classic ColBERT scoring step pools the candidate documents returned by a first-stage index, fetches their full embeddings, and re-scores them with the full MaxSim. [XTR](https://arxiv.org/abs/2605.00646) instead **scores documents directly from the per-token-hit scores** themselves: for each query token, a document gets credit for the maximum score it received in that token's top-`k_token` list, summed across query tokens, with the **minimum retrieved score** imputed for query tokens that did not surface any of its tokens (matching the original XTR paper). In PyLate, you can select between using ColBERT (full MaxSim scoring) or XTR (min score imputation) simply by using `retrieve.ColBERT` or `retrieve.XTR`
WARP implements this scoring natively, with aggressive IVF-based pruning that skips work which would not change the final ranking.
```

docs/index.md
```diff
@@ -328,7 +328,7 @@ queries = Dataset.from_list(mapping=queries)
 
 ## Retrieval
 
-PyLate provides an efficient index with [FastPLAID](https://github.com/lightonai/fast-plaid). Simply load a ColBERT model and initialize the index to perform retrieval.
+PyLate provides an efficient index with [FastPLAID](https://github.com/lightonai/fast-plaid). Simply load a ColBERT model and initialize the index to perform retrieval. A second backend, [WARP](https://github.com/pau-mensa/xtr-warp-rs), is also available via `indexes.WARP` (install with `pip install "pylate[warp]"`). WARP is not a drop-in for every model. Its centroid-pruning approximations can degrade recall on models that were not trained for them; see the [retrieval documentation](retrieval.md#xtr-retrieval) for the model compatibility notes and a concrete example.
```
NohTow:
```suggestion
PyLate provides an efficient index with [FastPLAID](https://github.com/lightonai/fast-plaid). Simply load a ColBERT model and initialize the index to perform retrieval.
```

docs/documentation/retrieval.md
robro612: > with aggressive IVF-based pruning that skips work which would not change the final ranking.

I think this implies a bit too much lossless pruning for what WARP's actually doing. It's still an approximation at the end of the day.

docs/index.md
NohTow, line 415:
```suggestion
Also note that a [WARP](https://github.com/pau-mensa/xtr-warp-rs) backend is also available via `indexes.WARP` (install with `pip install "pylate[warp]"`). WARP is an index to be used with [XTR-trained models](training.md#xtr-training) and might not work well with every model. It uses approximations that make retrieval faster and cheaper but can degrade performance on models that were not trained for them; see the [retrieval documentation](retrieval.md#xtr-retrieval) for the model compatibility notes and a concrete example.
```

docs/index.md
robro612: Doesn't this lose the info on WARP/how to install it? 

"Its centroid-pruning approximations can degrade recall on models that were not trained for them;" -> "It's missing score imputation, an aggressive approximation, can degrade retrieval quality on models not trained with the XTR objective;"

robro612: Ah, resolved below

PR #212, opened 2026-04-23
Add optional flash-maxsim backend to pylate.scores (up to 8× speedup, 40× memory reduction)

roipony:
# Add optional `flash-maxsim` backend to `pylate.scores`

## Summary

Adds a `backend=` kwarg (default `"auto"`) to the three public scoring
functions in `pylate/scores/scores.py`:

- `colbert_scores`
- `colbert_scores_pairwise`
- `colbert_kd_scores`

When CUDA is available and the optional `flash-maxsim` package is installed,
scoring is dispatched to fused Triton kernels that (a) avoid materialising
the full `[Nq, B, Lq, Ld]` score tensor and (b) replace the pairwise Python
`for` loop with a single varlen kernel call.

No change in behaviour for existing users — the new backend is opt-in via an
extras install and can be disabled per-call (`backend="torch"`) or globally
(`PYLATE_SCORES_BACKEND=torch`).

## Why

ColBERT-style scoring is bottlenecked by materialising `[Nq, B, Lq, Ld]`
just to take a max-then-sum. That's a lot of HBM traffic and (at training)
a lot of activation memory. `flash-maxsim` fuses the max+sum in SRAM and
saves only the per-query argmax in the backward, which cuts activation
memory by roughly the `Ld` factor and makes larger batch sizes fit on the
same GPU.

## Benchmarks — A100 80GB, fp16, flash-maxsim 0.2.1

Script: `benchmarks/bench_flash_backend.py`. All numbers are median of
3–10 runs after warm-up. `backend="torch"` is pylate's existing path;
`backend="flash"` is the new one.

### 1. Contrastive training step (fwd + bwd) — the headline

This is where the memory win changes what fits on one GPU.

| shape (Nq=B)                    | torch ms | flash ms | **speedup** | torch peak | flash peak | **mem ratio** |
|---------------------------------|---------:|---------:|------------:|-----------:|-----------:|--------------:|
| ColBERT  B=128  Lq=32  Ld=180   | 3.7      | 1.6      | **2.3×**    | 0.37 GB    | 0.07 GB    | 5.2×          |
| ColBERT  B=512  Lq=32  Ld=180   | 54.2     | 11.9     | **4.5×**    | 5.67 GB    | 0.67 GB    | 8.5×          |
| ColBERT  B=1K   Lq=32  Ld=180   | 215.9    | 46.4     | **4.7×**    | 22.57 GB   | 2.47 GB    | 9.1×          |
| ColBERT  B=2K   Lq=32  Ld=180   | **OOM**  | 192.2    | —           | **OOM**    | 9.52 GB    | **unlocked**  |
| ColPali  B=8   Lq=Ld=1024       | 2.5      | 2.0      | 1.2×        | 0.27 GB    | 0.03 GB    | 8.4×          |
| ColPali  B=16  Lq=Ld=1024       | 8.6      | 2.9      | **3.0×**    | 1.02 GB    | 0.06 GB    | 17.9×         |
| ColPali  B=32  Lq=Ld=1024       | 37.5     | 4.9      | **7.6×**    | 4.03 GB    | 0.13 GB    | 30.2×         |
| ColPali  B=64  Lq=Ld=1024       | 129.6    | 15.5     | **8.4×**    | 16.05 GB   | 0.39 GB    | **40.8×**     |

### 2. Contrastive inference (fwd only, no grad)

| shape                              | torch ms | flash ms | speedup |
|------------------------------------|---------:|---------:|--------:|
| B=32   Lq=32  Ld=180               | 0.20     | 0.18     | 1.1×    |
| B=128  Lq=32  Ld=180               | 1.55     | 0.41     | **3.8×** |
| B=256  Lq=32  Ld=180               | 5.92     | 1.20     | **4.9×** |
| B=64   Lq=128 Ld=512               | 3.47     | 0.56     | **6.2×** |
| B=32   Lq=1024 Ld=1024  (ColPali)  | 12.95    | 1.69     | **7.7×** |

### 3. Pairwise scoring (`colbert_scores_pairwise`)

Pylate's current path is a Python `for` loop, one pair at a time. Flash
packs the batch and runs a single varlen kernel.

| shape                   | torch ms | flash ms | **speedup** |
|-------------------------|---------:|---------:|------------:|
| B=256  text             | 19.8     | 1.1      | **18×**     |
| B=1000 text             | 77.6     | 3.5      | **22×**     |
| B=5000 text             | 382.7    | 16.7     | **23×**     |
| B=500  ColPali          | 42.5     | 3.4      | **12×**     |

## API

```python
from pylate.scores import colbert_scores

# Default: auto — flash if installed & CUDA, else torch. Non-breaking.
scores = colbert_scores(Q, D, q_mask, d_mask)

# Force the pure-torch path (e.g. for parity testing).
scores = colbert_scores(Q, D, q_mask, d_mask, backend="torch")

# Require flash; raises if flash-maxsim isn't installed or inputs aren't CUDA.
scores = colbert_scores(Q, D, q_mask, d_mask, backend="flash")
```

Env var `PYLATE_SCORES_BACKEND=torch|flash|auto` overrides the default
globally — useful for pinning a backend in CI.

Install:

```bash
pip install "pylate[flash]"
```

## Design notes

- **Non-breaking.** The new kwarg is optional, the extra is optional, and
  the auto path silently falls back to torch on CPU-only setups, on
  non-CUDA tensors, or if flash imports fail.
- **Graceful degradation.** If `backend="auto"` and the flash path raises
  at runtime, the wrapper falls back to torch. `backend="flash"` surfaces
  the exception instead, so users can debug.
- **Mask handling.** Pylate passes 0/1 masks of shape `[..., L]`; flash
  takes per-row lengths. Conversion is a single `sum(-1).to(int32)`.
- **Training.** When any input has `requires_grad`, the helper routes to
  `flash_maxsim_batched_train` (custom `autograd.Function`) so the backward
  stays fused. Only the argmax buffer `[Nq*B, Lq] int32` is saved instead
  of the full score tensor.
- **Pairwise.** `colbert_scores_pairwise` is a Python `for` loop today;
  the flash path packs per-pair shapes and does the whole batch in one
  `flash_maxsim_varlen` call.

## Tests

`tests/test_scores_flash.py` adds backend parity tests (skip-when-no-CUDA):

- `test_colbert_scores_parity` — with masks, fp16 + fp32
- `test_colbert_scores_pairwise_parity`
- `test_colbert_kd_scores_parity`
- `test_colbert_scores_grad_parity` — cos-sim on `grad_Q`/`grad_D` > 0.99
- `test_backend_arg_accepted_cpu` — default `"auto"` no-ops on CPU
- `test_backend_invalid` — bad `backend=` raises `ValueError`

All 9 parity tests pass on A100 80GB (max|Δ| ≤ 3e-3 on scores, grad cos-sim
= 1.00000 on grad_Q/grad_D).

## What's in the diff

- `pylate/scores/scores.py` — `backend` kwarg + `_resolve_backend` /
  `_try_flash` helpers; behaviour unchanged when `backend="torch"` or flash
  is unavailable.
- `pylate/scores/_flash_backend.py` — new lazy-imported helper that wraps
  flash-maxsim calls and handles mask→length conversion.
- `pyproject.toml` — new `[project.optional-dependencies]` entry
  `flash = ["flash-maxsim >= 0.2.1"]`.
- `tests/test_scores_flash.py` — new parity tests.
- `benchmarks/bench_flash_backend.py` — reproducible bench script.

No changes to `losses/*`, `rank/*`, or `evaluation/*`. They call the three
scoring functions directly and pick up the new default automatically.

## A note on KD (`colbert_kd_scores`)

KD is the smallest of the three wins. On realistic training shapes
(B=32–64, K=32–128, Lq=32, Ld=180) flash gives **1.1–1.5× speed** and
**slightly higher peak memory** (ratio ~0.8×). The invgrid CSR buffers
used in the backward dominate at KD shapes where the score tensor itself
is small (K instead of B×B rows).

We keep KD wired to the same `backend="auto"` path for consistency — it's
never a regression worth worrying about, and at larger Lq/K it should
grow. But the headline story is contrastive + pairwise, not KD.

| scenario                          | torch ms | flash ms | speedup |
|-----------------------------------|---------:|---------:|--------:|
| KD B=32  K=32   Lq=32  Ld=180     | 1.13     | 1.02     | 1.11×   |
| KD B=64  K=32   Lq=32  Ld=180     | 1.95     | 1.33     | 1.46×   |
| KD B=64  K=64   Lq=32  Ld=180     | 3.63     | 2.38     | 1.52×   |
| KD B=32  K=128  Lq=32  Ld=180     | 3.62     | 2.37     | 1.53×   |
| KD B=32  K=16   Lq=128 Ld=512     | 2.50     | 1.71     | 1.46×   |

## References

- `flash-maxsim` upstream: https://github.com/roipony/flash-maxsim
- v0.2.1 release notes: argmax init fix, invgrid backward, batched-train API
- Full benchmark data: `benchmarks/bench_flash_backend.py` (numbers above
  are A100 80GB; H100 results welcomed)


pylate/scores/_flash_backend.py
```diff
@@ -0,0 +1,160 @@
...
+def _mask_to_lengths(mask: Optional[torch.Tensor]) -> Optional[torch.Tensor]:
+    """Convert a [..., L] 0/1 mask to a [...] int32 length tensor. Returns
+    None if `mask` is None."""
+    if mask is None:
+        return None
+    # mask is typically FP16/FP32 with 0.0/1.0 values; sum along last dim.
+    return mask.sum(dim=-1).to(torch.int32).contiguous()
+
+
+def _inputs_supported(Q: torch.Tensor, D: torch.Tensor) -> bool:
```
NohTow, line 87: This is a nice check
Maybe a bit verbose but honestly it's well documented and avoid degenerated cases so it's fine to me

pylate/scores/_flash_backend.py
```diff
@@ -0,0 +1,160 @@
...
+    if not _inputs_supported(queries_embeddings, documents_embeddings):
+        raise RuntimeError("flash backend not applicable at this shape")
+
+    needs_grad = queries_embeddings.requires_grad or documents_embeddings.requires_grad
+
+    q_lens = _mask_to_lengths(queries_mask)
+    d_lens = _mask_to_lengths(documents_mask)
+
+    if needs_grad:
+        from flash_maxsim import flash_maxsim_batched_train
```
NohTow, line 118: @raphaelsty what do we think about importing in function against global import?

pylate/scores/_flash_backend.py
NohTow: Edit: When checking how it's used afterwards, I see two issues
The first one is that it tests for two things, whether it's _possible_ (cuda) and whether it's _useful_ (shapes)
In any case, it'll raise `raise RuntimeError("flash backend not applicable at this shape")`.

Shouldn't we rather:
- Make the cuda test, raise a RunTime error saying that it's only applicable to use for cuda inputs (and hint at using torch in those case)
- Make the shape test and yields a warning saying that it's probably suboptimal (but not crash)?

I think this can be done directly in the function here

pylate/scores/colbert.py
```diff
@@ -1,16 +1,59 @@
...
 
+import os
+
 import numpy as np
 import torch
 
 from ..utils.tensor import convert_to_tensor
 
+# Env override: PYLATE_SCORES_BACKEND={auto,torch,flash}
+_BACKEND_ENV = os.environ.get("PYLATE_SCORES_BACKEND", "auto").lower()
```
NohTow: @raphaelsty as discussed, this is one of the way of exposing it to the user when not directly using the scores (e.g, being the trainer during training, which is the main use case).
The other solution is leveraging a trainer kwargs, which is more HF-like, but at the same time everyone hates kwargs (and the usage of this should be minimal: anyone installing flashmaxsim probably wants to use it...)
cc @tomaarsen

pylate/scores/colbert.py
```diff
@@ -67,6 +115,22 @@ def colbert_scores(
     queries_embeddings = convert_to_tensor(queries_embeddings)
     documents_embeddings = convert_to_tensor(documents_embeddings)
 
+    resolved = _resolve_backend(backend)
+    if _try_flash(resolved, queries_embeddings, documents_embeddings):
+        try:
+            from ._flash_backend import colbert_scores_flash
```
NohTow, line 128: cc @raphaelsty, same question about in function import

Also I wonder if we could make it smoother

NohTow (commented):
Wonderful PR, thanks a lot for creating FlashMaxSim, making the PR and also rebasing due to us taking too much time to merge + adding new features in the mean time!

I've left a few comments in the code, mostly for @raphaelsty and I to discuss, probably not much for you to do on your sides, it's honestly an awesome functionality and nice implementations.

A few things I am flagging (some already are flagged within the code):
- There's this discussion around how to expose it to the user for the trainer (global variable against kwargs)
- I wonder if we should import at the top of the files or in function (there's two school of coding here I think)
- I wonder if we can make it a bit smoother for the call, like define the function based on the backend. Else we'll have a few if in all functions
- Thanks a lot for putting a nice bench with a lot of documentation, that's really helpful. I wonder if we should merge it though (we did not merge WARP benches)
- We might want to add it to the main documentation, I want people to know about it and use it!
- I wonder about the names of the additional dependency. Flash is often used for flash attention, but flashmaxsim might be verbose

pylate/scores/_flash_backend.py
raphaelsty: It's ok to import there, LGTM as it's an optional dependancy

pylate/scores/colbert.py
raphaelsty: I like it this way I think so, but I would move 

```python
_BACKEND_ENV = os.environ.get("PYLATE_SCORES_BACKEND", "auto").lower()
```

into:

```
def _resolve_backend(backend: str | None) -> str:
    """Resolve the effective backend from an explicit arg or the env override."""
    _BACKEND_ENV = os.environ.get("PYLATE_SCORES_BACKEND", "auto").lower()
    if backend not in ("auto", "torch", "flash"):
        raise ValueError(
            f"backend must be one of 'auto', 'torch', 'flash'; got {backend!r}"
        )
    return backend
```

It's the same but it enable to switch backend not only at import time

benchmarks/bench_flash_backend.py
```diff
@@ -0,0 +1,437 @@
+"""Benchmark the flash-maxsim backend vs the default torch path.
```
raphaelsty: You can remove the benchmarks folder, which has been useful for the mr review btw :)

PR #207, opened 2026-03-19
Add XTR retrieval and training support

robro612:
## Summary

This PR adds [XTR (conteXtual Token Retrieval)](https://research.google/pubs/rethinking-the-role-of-token-retrieval-in-multi-vector-retrieval/) support to PyLate, covering both retrieval and training.

### 1. XTR Retrieval

XTR retrieval performs per-token approximate nearest neighbor search followed by document-level scoring with min-imputation, as an alternative to ColBERT's full late-interaction reranking.

- **`pylate/retrieve/xtr.py`** — new `XTR` retriever class for XTR-style retrieval on Voyager/ScaNN indexes
- **`pylate/rank/rank.py`** — adds `score_xtr()` for document-level XTR scoring with min-imputation
- **`pylate/rank/__init__.py`** — exports `score_xtr`
- **`pylate/retrieve/__init__.py`** — exports `XTR` retriever
- **`pylate/retrieve/colbert.py`** — generalize type hint from `Voyager | PLAID` to `Base` index
- **Tests:** `tests/test_xtr_retriever.py`, `tests/test_xtr_scoring.py`

### 2. XTR Training

XTR training uses a token-level top-k scoring function with z-normalization instead of ColBERT's MaxSim. This requires seeing all documents at once (for global top-k), so both loss classes are updated to support a `requires_full_batch` score function.

- **`pylate/scores/scores.py`** — new `xtr_scores` and `xtr_kd_scores` functions, plus `XTRScores`/`XTRKDScores` callable classes for convenient default-k configuration. These classes set `requires_full_batch = True` as a class attribute, which the loss functions check to determine whether to pass the full document batch at once or chunk by group.
- **`pylate/scores/__init__.py`** — exports new scoring functions/classes
- **`pylate/losses/contrastive.py`** — detect `requires_full_batch` score metrics and pass all documents at once instead of chunking by group
- **`pylate/losses/cached_contrastive.py`** — same `requires_full_batch` support for cached contrastive training
- **`pylate/losses/distillation.py`** — add `temperature` parameter for XTR KD training
- **Tests:** `tests/test_xtr_scores.py`

PR #213, opened 2026-04-23
XTR training refactor

NohTow:
This PR builds upon #207, enhancing it to be better integrated into PyLate.
Essentially, the goal of the PR is to add XTR into PyLate, both training and retrieval (as a two-step retrieval mecanism, using either Voyager or SCANN, the "PLAID-version" of retrieval will be added in the form of WARP in #211).

The PR is a bit heavy but a lot of changes are tests (mostly Claude built, ngl) and I'll try to describe all the other changes.
- The first part is adding XTR scoring (for training, as they differ as the ones used for retrieval)). They are pretty similar to colbert scoring but with the additional masking. As they require some parameters, they are now defined as class rather than pure callable (but still define __call__, so still a callable). I believe we could maybe better re-use part of ColBERT into those (call ColBERT then mask), but it can be done later on.
- One of the biggest change is due to the fact that, unlike ColBERT, XTR *requires* all documents at training time to perform scoring (top-k filtering). To have an unified loss definition between ColBERT/XTR rather than two branches, I refactored the scoring part of contrastive loss to iterate over queries while feeding all documents at the scoring time.  We can still chunk the documents at scoring time (to avoid transient memory spike because of the big einsum) by defining document_chunk_size in the scoring function when defining the loss. We do not feed the whole set of queries at once to be able release some memory in GradCache when doing the backward pass. This is quite important, although we could also give the possibility to the user to use gradient checkpointing to avoid iterating at the cost of one additional pass (in addition to the gradcache one as well :/). This chunking dimension (over the query) is definable by the `scoring_micro_batch_size` value, so we can define the scoring chunking granularly. It also allowed to extends the behavior to the standard contrastive, which is cool because memory used during scoring was quite heavy (though I did not add document_chunk_size and kept the old per row iteration, we can change this later).

As discussed in DMs, I ran a bunch of tests and the training looks very healthy.
- Modified Voyager `__call__` output format to match the one of SCANN. There was a disrepency and the SCANN one was actually a better format because Voyager supposed all queries had the same lengths, which is less and less true with new models and would either crash or silently propagate mistakes. Only downside is if someone was using the output as a numpy array, but 1. they should not use it directly and 2. we are deprecating Voyager anyways.
- It adds XTR retrieve and associated scoring mechanism. Here I am still not satisfied entirely. I made the check over single stage and multiple stage indexes less brittle by adding a dedicated property, but I still think it is kind of brittle and will create odd things such as calling XTR retrieve on PLAID indexes returning ColBERT results. I think we can rather try to define a property saying what "algorithm" we are using (None for two stage, PLAID, XTR) and match it between index/retrieve function, but I am pretty sure we can do an even better integration. We can still merge somehow like this and deprecate XTR.retrieve later by just calling the new mechanism when calling this.


pylate/retrieve/xtr.py
```diff
@@ -0,0 +1,239 @@
...
+    ...     device="cpu",
+    ... )
+
+    >>> assert isinstance(results, list)
+    >>> assert len(results) == 1
+
+    """
+
+    def __init__(self, index: indexes.Base, verbose: bool = False) -> None:
+        if isinstance(index, indexes.PLAID):
```
NohTow: it'll break when we add WARP no?

pylate/indexes/__init__.py
```diff
@@ -1,7 +1,8 @@
 from __future__ import annotations
 
+from .base import Base
 from .plaid import PLAID
 from .scann import ScaNN
 from .voyager import Voyager
 
-__all__ = ["Voyager", "PLAID", "ScaNN"]
+__all__ = ["Base", "Voyager", "PLAID", "ScaNN"]
```
raphaelsty: Why do we want to expose Base ? Unless a good reason, better not to expose it

pylate/indexes/plaid.py
```diff
@@ -115,6 +115,8 @@ class PLAID(Base):
 
     """
 
+    is_end_to_end_index = True
```
raphaelsty, line 118: Why do we need this ? I'd like to avoid pattern like an other object calling an index.is_end_to_end_index such thing add debt. An other object from another module should not call attributes of indexes. Also I dislike class variable. It could be a property of the class if there really no choice and then update the abstract mother class but I'd rather like to avoid this

pylate/losses/cached_contrastive.py
```diff
@@ -259,76 +274,58 @@ def calculate_loss(self, reps, masks, with_backward: bool = False) -> Tensor:
...
-            labels = labels + rank * batch_size
-        losses: list[torch.Tensor] = []
         do_query_expansion = (
             self.model.do_query_expansion
             if hasattr(self.model, "do_query_expansion")
             else self.model.module.do_query_expansion
         )
+
+        N = len(embeddings_other)
+        docs_stacked = torch.stack(embeddings_other, dim=1)
```
raphaelsty, line 284: For claude, double check we need a torch.stack here and if we could not achieve the same thing with a torch.cat somehow. High propability stack is the best option, just check :)

pylate/retrieve/xtr.py
```diff
@@ -0,0 +1,178 @@
...
+    ...     device="cpu",
+    ... )
+
+    >>> assert isinstance(results, list)
+    >>> assert len(results) == 1
+
+    """
+
+    def __init__(self, index: indexes.Base, verbose: bool = False) -> None:
+        if index.is_end_to_end_index:
```
raphaelsty: I got it but there is a design issue here. Also you need a PhD in IR to understand the error. This is_end_to_end_index is my primary concern so far

pylate/retrieve/xtr.py
```diff
@@ -0,0 +1,178 @@
...
+import numpy as np
+import torch
+from tqdm.auto import tqdm
+
+from .. import indexes
+from ..rank import RerankResult, score_xtr
+from ..utils import iter_batch
+
+
+class XTR:
```
raphaelsty: The class is beautiful

pylate/scores/colbert.py
```diff
@@ -203,3 +203,223 @@ def colbert_kd_scores(
 
     scores = scores.max(axis=-1).values.sum(axis=-1)
     return scores
+
+
+class ColBERTScores:
```
raphaelsty, line 208: Please don't repeat the module name in the class name

```python
from pylate import scores

scores.ColBERT
```

is better than 

```python
from pylate import scores

scores.ColBERTScores # We get it that we are talking about scores, maybe we should add it multiple times like to make sure user understand

scores.ColBERTScoresThIsisTalkingAboutScoresYesScoresYouheardWellScores
```

I mean this is a big MR, so it's ok to joke a bit.

Here in the score module I expect a colbert.py file and an xtr.py file with a classs ColBERT and a class XTR. So please create two files from your additions.

pylate/scores/scores.py
```diff
@@ -203,3 +203,223 @@ def colbert_kd_scores(
...
+                documents_mask[:, j] if documents_mask is not None else None,
+            )
+            for j in range(N)
+        ]
+        # Stack to (Q_query, D, N) then flatten to (Q_query, D*N) with
+        # query-major ordering (doc d's k-th slot at column d*N + k).
+        return torch.stack(per_group, dim=2).reshape(-1, D * N)
+
+
+# Adapted from PrimeQA (https://github.com/primeqa/primeqa branch:xtr)
```
raphaelsty: Put this in the docstring please, there is a references field which the doc parse, otherwise no one will read this

pylate/scores/scores.py
```diff
@@ -203,3 +203,223 @@ def colbert_kd_scores(
...
+
+        if queries_mask is not None:
+            topk_scores_max = topk_scores_max * queries_mask.unsqueeze(-1)
+
+        scores_sum = topk_scores_max.sum(dim=1)  # (Qb, Db)
+        Z = topk_scores_max.gt(0).float().sum(dim=1).clamp_(min=1e-3)
+        return (scores_sum / Z).float()
+
+
+class XTRKDScores(XTRScores):
```
raphaelsty: KDXTR would be better naming I think so, put this in xtr.py within scores module.

raphaelsty (commented):
Amazing MR, this is a very cool feature, I leaved some comments, tried to be funny, open for discussion ofc. I think overall the MR missed something which is telling peoples how to use this index and when to use it and with which model appart from the comments I made :)

pylate/indexes/__init__.py
NohTow: I think it comes from @robro612 
no strong opinion if it's not needed

robro612: I think I exposed it because I wanted to generalize the type hints in the retriever class. I think that could be done hidden without exposing it in the api, or just omitting it.

pylate/indexes/plaid.py
NohTow: We discussed this and we can still discuss it. I would rather have a proper attribute than typechecking in Python, which is brittle and already broke before.

pylate/retrieve/xtr.py
NohTow: let's call after to discuss it

pylate/retrieve/xtr.py
NohTow: Used to be much worse, thanks!

pylate/indexes/plaid.py
robro612: @NohTow at one point suggested the index class structure could be split into E2E/DocumentIndex and TokenIndex to distinguish the WARP/PLAID from regular old token-level indices like ScaNN. Then this check becomes an `isinstance` check, which is kind of what it already is in the retriever class. 

Bottom line: do you consider a class attribute to be meaningfully less desirable than a class itself?

pylate/scores/scores.py
raphaelsty: Here is an exemple of numpy-style docstring :) 
```
  References
  ----------
  - [NanoBEIR](https://huggingface.co/collections/zeta-alpha-ai/nanobeir-66e1a0a
  f21dfd93e620cd9f6)
```

pylate/indexes/__init__.py
NohTow: Solved for both indexes and retrieve

pylate/indexes/plaid.py
NohTow: chatted it through, we go for an abstract property and will add another one for is_xtr/is_colbert later on.
It's a bit verbose imho but honestly not a big deal + it's clearer/safer. we shouldn't have too many "retrieval type", else we'll need a big refactor anyways so let's ship

pylate/scores/colbert.py
NohTow: We discussed it and although this is a very fair comment, we have had scores.colbert_scores for a long time and refactoring all of this would be scope creep.
open an issue/another pr if needed

pylate/scores/scores.py
NohTow: should be good now

pylate/scores/scores.py
NohTow: splitted the scores, did not rename, I honestly do not mind

raphaelsty:
LGTM you can merge

NohTow:
Thanks @robro612 @arn4 @raphaelsty, good job!

PR #223, opened 2026-05-24
Add TachiomIndex and return_token_ids for ColBERT.encode

SilvioM97:
# PyLate Fork Changes

Changes made to the upstream [PyLate](https://github.com/lightonai/pylate) codebase for the TACHIOM integration. All changes are purely additive — no existing behaviour was modified.

---

## `pylate/models/colbert.py` — `ColBERT.encode()`: add `return_token_ids` parameter

**Why:** `TachiomIndex.add_documents` needs the vocabulary token IDs (BERT token IDs) of each kept token alongside the embeddings, in order to pass them to `Tachiom.build_from_arrays()`. TAC (Token-Aware Clustering) groups tokens by their vocabulary ID and runs a separate k-means per group; without real token IDs the index degrades to a single global k-means (no per-type clustering).

**What changed:**
- New parameter `return_token_ids: bool = False` on `encode()`.
- When `True`, the return value is a tuple `(embeddings, token_ids)` where `token_ids` is a `list[np.ndarray[uint32]]` — one array per document/query, aligned 1-to-1 with the corresponding embedding array (tokens align after masking).
- Token IDs are extracted from `features["input_ids"]` using the same mask that selects which token embeddings survive (skiplist + attention_mask for documents; attention_mask or all-ones for queries with/without expansion).
- Raises `ValueError` if combined with `pool_factor > 1` (pooling merges tokens so IDs no longer align with embeddings).
- No behaviour change when `return_token_ids=False` (default).

---

## `pylate/indexes/__init__.py` — export `TachiomIndex`

**Why:** provide TACHIOM as a drop-in `Base`-compatible index backend for PyLate, enabling TAC + PQ + HNSW retrieval for ColBERT-style multi-vector search.

**Interface:**
```python
index = TachiomIndex(index_folder="my_indexes_dir", index_name="my_index_name")
# All build and search parameters are optional with corpus-adaptive defaults.
# See the class docstring or TACHIOM_FOR_PYLATE.md for the full parameter reference.

embeddings, token_ids = model.encode(docs, is_query=False, return_token_ids=True)
index.add_documents(doc_ids, embeddings, documents_token_ids=token_ids)

results = index(query_embeddings, k=10)  # list[list[RerankResult]]
```

**Design decisions:**

- `is_end_to_end_index = True` — `__call__` returns final ranked results directly, no external reranking step needed.

- `add_documents` accepts an optional `documents_token_ids` kwarg (list of uint32 arrays from `encode(..., return_token_ids=True)`).
  - If omitted: all tokens are assigned ID 0, so TAC degrades to a single global k-means over all vectors. This is **significantly slower** than TAC and reduces retrieval quality. A `UserWarning` is issued.
  - On the first call: a `logging.warning` informs the user that the index is built once and subsequent calls will have no effect. The `is_indexed` guard fires at the top of the method, before any array conversion, so no-op calls are cheap.
  - On subsequent calls: a `UserWarning` is issued and the call is a no-op.

- `remove_documents`: issues a `UserWarning` and is a no-op.

- `get_documents_embeddings`: returns approximate token embeddings reconstructed from coarse centroids and PQ codes via `approx = coarse_centroid + norm × PQ_residual`. Implemented through a new `get_document_embeddings(doc_id)` binding in Tachiom (`python.rs`) that calls `residuals.encoder().decode_vector()` via the `MultiVecEncoder` trait. Returns `list[list[np.ndarray]]` matching the input nesting; each array is `(n_tokens, dim)` float32. Raises `ValueError` for out-of-range IDs.

- `__call__` supports any input shape: list of 2D arrays, a single 2D array (treated as one query), or a 3D array `(n_queries, n_tokens, dim)`. Ragged query lengths are handled via the `offsets` parameter of `batch_search`; uniform lengths use the faster `offsets=None` path.

- Document ID mappings (`_doc_id_to_int` / `_int_to_doc_id`) are cached in memory after the first load, so disk reads happen at most once per instance lifetime.

- After `add_documents` completes, the three auto-derived parameters (`total_centroids`, `tac_micro_threshold`, `tac_small_threshold`) are written back to their instance variables, replacing the user-supplied `None`. This makes `__repr__` display the actual values used rather than `None`.

- `ef_search` defaults to `None` in `TachiomIndex.__init__`. At search time, tachiom resolves `None` to `round(1.5 × k_centroids)`, keeping the two coupled automatically. Pass an explicit value only to deviate from the 1.5× rule.

- `__repr__` (indexed state) prints the full build and search configuration — all hyperparameters grouped into `— build —` and `— search —` sections — so `print(index)` after building gives a complete record of exactly what was used. `ef_search` is shown as the resolved effective value with an `(auto)` suffix when using the default.

- Sentinel filtering: Tachiom pads short result lists with `doc_id=u32::MAX`, `score=-inf`. These are stripped before returning results.

**On-disk layout** under `my_indexes_dir/my_index_name/` (i.e. `{index_folder}/{index_name}/`):

| File | Contents |
|---|---|
| `tachiom.bin` | Serialised Tachiom index (bincode) |
| `doc_id_to_int.pkl` | `str → int` document ID mapping |
| `int_to_doc_id.pkl` | `int → str` document ID mapping |

---

### 1. Incremental `add_documents` and `remove_documents`

**Status:** currently not supported. Both are no-ops with a `UserWarning`.

**What's needed (Tachiom/vectorium side — major effort):**
- Dynamic insertion and deletion requires making vectorium's datasets and HNSW graph mutable at runtime. Currently both are built once and serialised as immutable structures.

**Workaround:** call `add_documents` with the full desired corpus each time (the index is built from scratch on the first call).

---

### 2. Subset search

**Status:** currently not supported. WARP's `__call__` accepts a `subset` parameter that restricts search to a specified list of document IDs. Tachiom's `batch_search` has no equivalent.


pylate/indexes/tachiom.py
```diff
@@ -0,0 +1,466 @@
...
+        beta: int | None = None,
+        lambda_: float | None = None,
+        num_threads: int = 0,
+    ) -> None:
+        try:
+            from tachiom import Tachiom as _Tachiom
+            from tachiom import auto_build_params as _auto_build_params
+        except ImportError:
+            raise ImportError(
+                "tachiom is not installed. Please install it with: `pip install tachiom`."
```
NohTow, line 164: we probably want to add it as (at least) an optional dependency in the project so people can do pip install pylate[tachiom].
also cc @raphaelsty, considering the lightweight dependencies of their report, we can maybe add it in the default?

pylate/indexes/tachiom.py
```diff
@@ -0,0 +1,466 @@
...
+        self._ensure_mappings()
+        return [
+            [
+                self._index.get_document_embeddings(self._doc_id_to_int[doc_id])
+                for doc_id in group
+            ]
+            for group in documents_ids
+        ]
+
+    def __repr__(self) -> str:
```
NohTow, line 513: I wonder about this
I'm not against having a function to prob the parameters used, but maybe we should return a dict? (and maybe add it for all indexes).
It feels odd in the current codebase maybe

pylate/indexes/tachiom.py
```diff
@@ -0,0 +1,466 @@
...
+        if self.is_indexed:
+            warnings.warn(
+                "TachiomIndex has already been built and does not currently support "
+                "incremental updates. This call has no effect.",
+                UserWarning,
+                stacklevel=2,
+            )
+            return self
+
+        logger.warning(
```
NohTow: I wonder if we should have this warning at any insert
I get the point of wanting to share the information but imho this could be in documentation + the user will know when it crashes next time (I'm just thinking it might be verbose)

pylate/indexes/tachiom.py
```diff
@@ -0,0 +1,466 @@
...
+            "TachiomIndex is built once from all documents. "
+            "Subsequent calls to add_documents will have no effect — "
+            "pass all documents you want indexed in this call."
+        )
+
+        embeddings_f32 = [self._to_f32(e) for e in documents_embeddings]
+        new_doclens = np.array([e.shape[0] for e in embeddings_f32], dtype=np.int32)
+        new_vectors_u16 = (
```
NohTow: naive question: why are we casting to f32 to then cast to u16?
also, why aren't we modifying in place? this uses more memory (but maybe doclens isn't an user facing variable anyways)

pylate/indexes/tachiom.py
```diff
@@ -0,0 +1,466 @@
...
+        self._int_to_doc_id: dict | None = None
+        self.is_indexed = os.path.exists(self._tachiom_path)
+        if self.is_indexed:
+            self._index = self._Tachiom.load(self._tachiom_path)
+
+    def _ensure_loaded(self) -> None:
+        if self._index is None:
+            self._index = self._Tachiom.load(self._tachiom_path)
+
+    def _ensure_mappings(self) -> None:
```
NohTow, line 213: FWIW, mapping are only set when _saving_ the ids
Maybe i'm being naive but if everything was loaded at init (not just the path), we won't need any ensure?
Expect on the is_indexed when trying to search/get embeddings

pylate/indexes/tachiom.py
```diff
@@ -0,0 +1,466 @@
...
+            tac_small_threshold=self.tac_small_threshold,
+        )
+        total_centroids = _p["total_centroids"]
+        micro_threshold = _p["tac_micro_threshold"]
+        small_threshold = _p["tac_small_threshold"]
+        # Store resolved values so __repr__ can show the actual config used.
+        self.total_centroids = total_centroids
+        self.tac_micro_threshold = micro_threshold
+        self.tac_small_threshold = small_threshold
+        logger.info(
```
NohTow: I wonder if we should let this, might be verbose + not very consistent with the other indexes (but it's also kind of useful to debug at times)

pylate/indexes/tachiom.py
```diff
@@ -0,0 +1,466 @@
...
+                stacklevel=2,
+            )
+            new_token_ids = np.zeros(int(new_doclens.sum()), dtype=np.uint32)
+
+        doc_id_to_int = {doc_id: i for i, doc_id in enumerate(documents_ids)}
+        int_to_doc_id = {i: doc_id for i, doc_id in enumerate(documents_ids)}
+        self._save_mappings(doc_id_to_int, int_to_doc_id)
+
+        token_ids_cont = np.ascontiguousarray(new_token_ids, dtype=np.uint32)
+        _p = self._auto_build_params(
```
NohTow: maybe `_auto_param` is more readable?

pylate/indexes/tachiom.py
```diff
@@ -0,0 +1,466 @@
...
+        self.total_centroids = total_centroids
+        self.tac_micro_threshold = micro_threshold
+        self.tac_small_threshold = small_threshold
+        logger.info(
+            "TachiomIndex: %d docs, %d tokens, dim=%d — "
+            "total_centroids=%d, micro_threshold=%d, small_threshold=%d",
+            len(new_doclens), len(new_token_ids), embeddings_f32[0].shape[1],
+            total_centroids, micro_threshold, small_threshold,
+        )
+        self._index = self._Tachiom.build_from_arrays(
```
NohTow, line 367: thanks for adding this functionality, much cleaner than saving/loading to/from disk!

pylate/indexes/tachiom.py
```diff
@@ -0,0 +1,466 @@
...
+        Parameters
+        ----------
+        queries_embeddings
+            Query token embeddings. Accepts a list of 2D arrays
+            ``(n_tokens, dim)``, a single 2D array (treated as one query),
+            or a 3D array ``(n_queries, n_tokens, dim)``.
+        k
+            Number of results to return per query.
+        """
+        if not self.is_indexed:
```
NohTow, line 407: imho this is the only check that's needed no?

pylate/indexes/tachiom.py
```diff
@@ -0,0 +1,466 @@
...
+        lens = [q.shape[0] for q in queries_list]
+        tokens = np.ascontiguousarray(np.vstack(queries_list), dtype=np.float32)
+
+        if len(set(lens)) == 1:
+            offsets = None
+        else:
+            offsets = np.zeros(n_queries + 1, dtype=np.uint64)
+            np.cumsum(lens, out=offsets[1:])
+
+        scores, doc_ids = self._index.batch_search(
```
NohTow, line 441: maybe it's dumb because tachiom is so efficient but should we expose any bs param?

pylate/indexes/tachiom.py
```diff
@@ -0,0 +1,466 @@
...
+            alpha=self.alpha,
+            beta=self.beta,
+            lambda_=self.lambda_,
+        )
+
+        results = []
+        for query_scores, query_doc_ids in zip(scores, doc_ids):
+            query_results = []
+            for score, doc_id in zip(query_scores, query_doc_ids):
+                if doc_id == _SENTINEL_DOC_ID:
```
NohTow, line 459: note for myself: check

pylate/models/colbert.py
```diff
@@ -554,8 +555,19 @@ def encode(
...
         protected_tokens
             The number of tokens at the beginning of the sequence that should not be pooled. Defaults to 1 (CLS token).
+        return_token_ids
+            If True, also returns the vocabulary token IDs of the kept tokens as a list of uint32 numpy arrays (one per
+            document/query). Each array aligns 1-to-1 with the token embeddings for that document after masking
+            (skiplist + attention mask). When True, the return value is a tuple ``(embeddings, token_ids)``.
+            Defaults to False.
 
         """
+        if return_token_ids and pool_factor > 1:
```
NohTow: Good catch!
Dumb (research) question: don't you think we could extend the ids from the original ones?

pylate/models/colbert.py
```diff
@@ -652,6 +664,7 @@ def encode(
         self.to(device)
 
         all_embeddings = []
+        all_token_ids: list[torch.Tensor] = []
```
NohTow: note for myself: check that the tokenizer always returns tensor in this path

pylate/models/colbert.py
```diff
@@ -800,6 +819,11 @@ def encode(
                 for embedding in all_embeddings
             ]
 
+        if return_token_ids:
```
NohTow: What happens when we use convert_to_tensor/convert_to_numpy?

pylate/models/colbert.py
```diff
@@ -800,6 +819,11 @@ def encode(
...
+            token_ids_out = [ids.numpy().astype(np.uint32) for ids in all_token_ids]
+            if input_was_string:
```
NohTow: I'm not sure to get this one

NohTow (commented):
Amazing PR, very much looking forward to get TACHIOM in PyLate!
I've added a few comments, sorry if some are naive, wanted to get a first pass ASAP.

I think this is already in a very good state but I've some nits to make it as clean as possible.

Also cc @tomaarsen, I wanted to ask you what you think is better to get the `token_ids` of the embeddings. I think having a param for the encode function is very fine, but maybe you prefer to require people to retokenize after the encode?

pylate/indexes/tachiom.py
SilvioM97: added as optional (pip install pylate[tachiom]). Happy to promote it to default if @raphaelsty thinks it makes sense.

pylate/indexes/tachiom.py
SilvioM97: Good point. We can add a get_config() method to TachiomIndex that returns a dict of all parameters, and simplify __repr__ to just path + stats. Should we do that?

pylate/indexes/tachiom.py
SilvioM97: Right, removed it

pylate/indexes/tachiom.py
SilvioM97: The intermediate f32 is unavoidable, _to_f32() normalizes all input types (tensors, f16, f32) before converting to f16 bit patterns. It's a temporary copy only during the build call to give tachiom the format it needs.

pylate/indexes/tachiom.py
SilvioM97: Fix: call the ensure methods at init

pylate/indexes/tachiom.py
SilvioM97: Fix: change logger.info -> logger.debug (still available for debugging but not shown by default)

pylate/indexes/tachiom.py
SilvioM97: Fix: renamed to params

pylate/indexes/tachiom.py
SilvioM97: Correct, handled by the init fix above. The ensure calls in __call__ are now no-ops, kept as safety guards consistent with WARP's pattern

pylate/indexes/tachiom.py
SilvioM97: TACHIOM’s batch_search already handles everything in parallel internally, so it shouldn’t be necessary

pylate/models/colbert.py
SilvioM97: Agreed, it should be feasible (e.g.  assign each pooled output the most frequent vocabulary ID from its pool). We've kept the ValueError for safety since it would need proper testing to validate retrieval quality. Happy to tackle this as a follow-up.

pylate/models/colbert.py
SilvioM97: The embeddings part of the tuple follows the flags as normal, numpy if convert_to_numpy=True, tensors if convert_to_tensor=True. The token IDs are always numpy uint32 regardless. The only edge case is convert_to_tensor=True, return_token_ids=True, which returns mixed types (tensor embeddings + numpy IDs), but it's not a problem for TACHIOM (which is the only reason we use token ids now)

pylate/models/colbert.py
SilvioM97: PyTorch stores token IDs as int64 tensors, but tachiom's build_from_arrays specifically needs uint32. So we convert. Added a one-line comment in the code.

pylate/indexes/tachiom.py
NohTow: LGTM, but do we actually need it?

SilvioM97: Probably not strictly needed for now. Happy to keep __repr__ as-is and add get_config() later if it comes up.

pylate/indexes/tachiom.py
robro612: Given that ColBERT does some token-level postprocessing (e.g. skiplist), I think it would be good to keep the id handling alongside the encode method. 

having paired ids will be useful beyond TACHIOM too!

pylate/models/colbert.py
```diff
@@ -555,18 +555,22 @@ def encode(
...
+            per-document token embeddings — the existing behaviour.
+            ``None``: a list of per-document dicts with keys ``"token_embeddings"``
+            (shape ``[seq_len, dim]``), ``"input_ids"``,
+            ``"attention_mask"``, and ``"masks"`` (combined skiplist + attention mask,
+            bool). The caller applies the mask to obtain aligned embeddings and token
+            IDs. Incompatible with ``pool_factor > 1``.
 
         """
-        if return_token_ids and pool_factor > 1:
+        if output_value not in ("token_embeddings", None):
```
robro612, line 568: This seems like a more opaque interface than the additive flag `return_token_ids`. Could you explain your reasoning for changing it?

SilvioM97: You're right that `return_token_ids` is more self-explanatory in isolation. We changed it specifically for ST compatibility: `output_value` already exists in `SentenceTransformer.encode()` with the same semantics, and ColBERT inherits from it. Using a TACHIOM-specific `return_token_ids` flag would diverge from the `SentenceTransformer.encode()` interface that ColBERT already inherits from. The change was discussed with @NohTow for this reason.
Happy to discuss it and find alternative solutions!

NohTow: I asked @tomaarsen what would be the cleanest way of returning the token ids and he answered back with the mecanism that is currently in ST and could be extended to other type of returns as well
It's a bit less direct for what we are doing, but is more extendable

pylate/models/colbert.py
```diff
@@ -554,8 +555,23 @@ def encode(
...
+            ``None``: a list of per-document dicts with keys ``"token_embeddings"``
+            (shape ``[seq_len, dim]``), ``"input_ids"``,
+            ``"attention_mask"``, and ``"masks"`` (combined skiplist + attention mask,
+            bool). The caller applies the mask to obtain aligned embeddings and token
+            IDs. Incompatible with ``pool_factor > 1``.
 
         """
+        if output_value not in ("token_embeddings", None):
+            raise ValueError(
+                f"output_value must be 'token_embeddings' or None, got {output_value!r}."
```
NohTow, line 570: I think here we would benefit from a comment/more explicit error saying that it would be ill-defined to return ids while we merged some. That being said, I wonder if we should _prevent_ it or raise a warning.
We are currently returning all ids, even the ones that do not have corresponding embeddings anyways... But maybe that's for another issue, let's move on (let's just add a comment about it)

SilvioM97: Added a comment above the check explaining the reasoning. On prevent vs. warn: I'd keep the hard error, a warning is easy to miss, and a caller who proceeds anyway would silently get misaligned input_ids/token_embeddings, which is worse than failing at encode time imho. We can surely address this later!

